PR #3, opened 2016-09-09
New: Mark TypeScript-specific identifiers as used (fixes #2)

### LINKED ISSUE #2: Manually mark TypeScript-specific nodes as used
From @nzakas:

> In the short-term, we can probably add a rule in eslint-plugin-typescript that manually marks the identifier as used. eslint-plugin-react had to do something similar for some JSX stuff: https://github.com/yannickcr/eslint-plugin-react/blob/master/lib/rules/jsx-uses-vars.js

Example of decorators can be seen here: https://github.com/eslint/typescript-eslint-parser/issues/55

JamesHenry:
**Not yet ready to be merged**

- [x] Class implements interface 
- [x] All decorator variants
- [x] Variable declaration type annotation
- [ ] All function returnTypes
- [ ] Class properties defined using `private` or `public` on parameters in constructor

### DIFF
--- a/lib/rules/typescript-uses-vars.js
+++ b/lib/rules/typescript-uses-vars.js
@@ -1,0 +1,154 @@
+/**
+ * @fileoverview Prevent TypeScript-specific variables being falsely marked as unused
+ * @author James Henry
+ */
+"use strict";
+
+/**
+ * Record that a particular variable has been used in code
+ *
+ * @param {Object} context The current rule context.
+ * @param {String} name The name of the variable to mark as used.
+ * @returns {Boolean} True if the variable was found and marked as used, false if not.
+ */
+function markVariableAsUsed(context, name) {
+    var scope = context.getScope();
+    var variables;
+    var i;
+    var len;
+    var found = false;
+
+    // Special Node.js scope means we need to start one level deeper
+    if (scope.type === "global") {
+        while (scope.childScopes.length) {
+            scope = scope.childScopes[0];
+        }
+    }
+
+    do {
+        variables = scope.variables;
+        for (i = 0, len = variables.length; i < len; i++) {
+            if (variables[i].name === name) {
+                variables[i].eslintUsed = true;
+                found = true;
+            }
+        }
+        scope = scope.upper;
+    } while (scope);
+
+    return found;
+}
+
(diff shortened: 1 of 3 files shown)


lib/rules/typescript-uses-vars.js
```diff
@@ -0,0 +1,154 @@
...
+ * @fileoverview Prevent TypeScript-specific variables being falsely marked as unused
+ * @author James Henry
+ */
+"use strict";
+
+/**
+ * Record that a particular variable has been used in code
+ *
+ * @param {Object} context The current rule context.
+ * @param {String} name The name of the variable to mark as used.
```
nzakas: lowercase "string" and "boolean"

nzakas:
FYI: https://github.com/eslint/eslint/issues/7129

JamesHenry:
Great to see! That should remove the need for some, but not all, of the points covered by this plugin rule

nzakas:
It should remove the need to check type annotations, function return types, and decorators here. The others would still be necessary.

lib/rules/typescript-uses-vars.js
```diff
@@ -0,0 +1,154 @@
+/**
```
nzakas: I don't think we need the word "typescript" in the rule, since the plugin is already called "typescript". Maybe just `no-unused-vars` to match the name of the core rule?

JamesHenry: I based that on the `jsx-uses-vars` "rule" in the react plugin.

I am putting rule in quotes because it is not really a rule at all. All the code is doing is marking things as used; it doesn't define a set of behaviour that the source code must implement, like the `no-unused-vars` core rule does.

I am totally happy to go with your final decision on the name, but that was the reasoning :)

nzakas: I still think `no-unused-vars` makes more sense because it will trigger warnings from the core rule of the same name.

JamesHenry:
Given that some of these are low-hanging fruit, whereas others are being blocked by much broader topics (e.g. escope dependency), I am going to break this up into an initial `no-unused-vars` PR which will only cover the non-blocked stuff.

PR #6, opened 2016-10-04
New: Add rule no-triple-slash-reference

dannyfritz:
Rule based off of TSLint's [no-reference rule](http://palantir.github.io/tslint/rules/no-reference/)

### DIFF
--- a/docs/rules/no-triple-slash-reference.md
+++ b/docs/rules/no-triple-slash-reference.md
@@ -1,0 +1,41 @@
+# Enforces reference comments are not used (no-triple-slash-reference)
+
+Triple-slash reference directive comments should not be used anymore. Use `import` instead.
+
+Before TypeScript adopted ES6 Module syntax,
+triple-slash reference directives were used to specify dependencies.
+Now that we have `import`, triple-slash reference directives are discouraged for specifying dependencies
+in favor of `import`.
+
+A triple-slash reference directive is a comment beginning with three slashes followed by a path to the module being imported:
+`/// <reference path="./Animal" />`.
+ES6 Modules handle this now:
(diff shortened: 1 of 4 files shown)


docs/rules/no-triple-slash-reference.md
```diff
@@ -0,0 +1,31 @@
+# Enforces reference comments are not used (no-triple-slash-reference)
+
+Reference comments, `/// <reference />`, should not be used anymore. Use `import` instead.
```
nzakas: Can you explain this a bit more? As someone unfamiliar with TypeScript, I think more context would be helpful.

tests/lib/rules/no-triple-slash-reference.js
```diff
@@ -0,0 +1,45 @@
...
+
+//------------------------------------------------------------------------------
+// Tests
+//------------------------------------------------------------------------------
+
+var ruleTester = new RuleTester();
+ruleTester.run("no-triple-slash-reference", rule, {
+
+    valid: [
+        {
```
nzakas: Can you add a valid test that uses comments? I'd just like to be sure the rule doesn't incorrectly warn on double-slash, for example.

tests/lib/rules/no-triple-slash-reference.js
```diff
@@ -0,0 +1,45 @@
...
+            code: "import Animal from \"Animal\"",
```
nzakas: This really isn't a useful test because the rule isn't looking for import statements.

dannyfritz:
Requested changes have been made.

nzakas:
Thanks!

PR #8, opened 2016-10-13
New: Add rule no-any

dannyfritz:
Rule based off of TSLint's [no-any](https://palantir.github.io/tslint/rules/no-any/) rule


JamesHenry:
Thanks for your continued efforts, @dannyfritz!

Could we go for `no-explicit-any` as the name of the rule?

I think it is a better explanation of what is actually being checked, and that was it is more clearly a complement to the TypeScript compiler flag `noImplicitAny`.

Could you also briefly touch on that relationship in your docs?

dannyfritz:
Rebased and made changes. :)

JamesHenry:
LGTM, thanks!

PR #9, opened 2016-10-13
New: Mark interfaces and decorators as used (refs #2)

JamesHenry:
Custom type annotations and return types being marked as used will be enabled via the new ESLint escope fork


nzakas:
LGTM

PR #14, opened 2017-01-01
Fix: incorrect type-annotation-spacing test

weirdpattern:
Fixes one of the tests related to rule 'type-annotation-spacing'

### DIFF
--- a/tests/lib/rules/type-annotation-spacing.js
+++ b/tests/lib/rules/type-annotation-spacing.js
@@ -29,7 +29,7 @@ ruleTester.run("type-annotation-spacing", rule, {
             parser: "typescript-eslint-parser"
         },
         {
-            code: "function foo(a:string) {}",
+            code: "function foo(a: string) {}",
             parser: "typescript-eslint-parser"
         }
     ],


JamesHenry:
The fact that this was _not_ causing the build to fail before is the real issue that needs to be addressed.

It might be related to the older version of the `typescript-eslint-parser` that this project is currently requiring, please can you look into that?

weirdpattern:
Of course, I will work on this today

weirdpattern:
@JamesHenry this is definitely the parser.

0.4.0 wasn't generating typeAnnotation nodes for function arguments. 
This was addressed [here](https://github.com/eslint/typescript-eslint-parser/pull/112/commits/b11e1e6f72f826f372d5b095af0277910b139804)  

Shall I use this PR to update the package.json dependency too? 

TypeScript Parser 0.4.0
node-tsp-0 4 0

TypeScript Parser 1.0.0
node-tsp-1 0 0

nzakas:
We should update the parser in a separate OR.

weirdpattern:
In that case, this PR is ready
I will submit a new PR with the typescript-eslint-parser version change

Issue #5, opened 2016-09-12
[META] TSLint Compatibility


dannyfritz:
- [x] [array-type](https://palantir.github.io/tslint/rules/array-type/) (#211 )
- [x] [adjacent-overload-signatures](http://palantir.github.io/tslint/rules/adjacent-overload-signatures/) (#37 )
- [x] [ban-types](https://palantir.github.io/tslint/rules/ban-types/) (#181 )
- [ ] [callable-types](https://palantir.github.io/tslint/rules/callable-types/)
- [x] [class-name](https://palantir.github.io/tslint/rules/class-name/) (#44 `class-name-casing`)
- [x] [interface-name](http://palantir.github.io/tslint/rules/interface-name/) (#10 `interface-name-prefix`)
- [x] [interface-over-type-literal](https://palantir.github.io/tslint/rules/interface-over-type-literal/) (#15 `no-type-literal`)
- [x] [member-access](http://palantir.github.io/tslint/rules/member-access/) (#4 `explicit-member-accessibility`)
- [x] [member-ordering](http://palantir.github.io/tslint/rules/member-ordering/) (#23 )
- [x] [no-angle-bracket-type-assertion](http://palantir.github.io/tslint/rules/no-angle-bracket-type-assertion/) (#13 )
- [x] [no-any](http://palantir.github.io/tslint/rules/no-any/) (#8 `no-explicit-any`)
- [x] [no-empty-interface](https://palantir.github.io/tslint/rules/no-empty-interface/) (#39 )
- [ ] [no-inferred-empty-object-type](https://palantir.github.io/tslint/rules/no-inferred-empty-object-type/)
- [x] [no-inferrable-types](http://palantir.github.io/tslint/rules/no-inferrable-types/) (#124 )
- [x] [no-internal-module](http://palantir.github.io/tslint/rules/no-internal-module/) (#22 `prefer-namespace-keyword`)
- [ ] [no-mergeable-namespace](http://palantir.github.io/tslint/rules/no-mergeable-namespace/)
- [ ] [no-misused-new](https://palantir.github.io/tslint/rules/no-misused-new/)
- [x] [no-namespace](http://palantir.github.io/tslint/rules/no-namespace/) (#21 )
- [x] [no-non-null-assertion](https://palantir.github.io/tslint/rules/no-non-null-assertion/) (#108 )
[...]

PR #15, opened 2017-01-05
New: Add rule no-type-literal

weirdpattern:
This includes

New rule no-type-literal
Corresponding test
Corresponding documentation
Update to README.md to include the new rule


weirdpattern:
Hi @dannyfritz, could you please reference this PR from #5 
Also, please let me know if you need a hand with the some of these rules, I'll be glad to help!

dannyfritz:
Which TSLint does this correlate to?

docs/rules/no-type-literal.md
```diff
@@ -0,0 +1,121 @@
+# Enforces the use of interfaces over type literals (`type T = {...}`).
```
JamesHenry: TypeScript refers to these as "Type Aliases" in their docs, so please could we go for that name

docs/rules/no-type-literal.md
```diff
@@ -0,0 +1,121 @@
...
+make interfaces a better choice:   
+- Interfaces can be extended or implemented by other types.
+- Interfaces create a new name that is used everywhere, so when debugging is easier to identify 
+the base type of an object.
+
+## Rule Details
+
+This rule aims to standardise the use of interfaces across the codebase.
+
+## Options
```
JamesHenry: As with the other PR, please can we stick to the exceptions being `false` by default for "no-" rules, I feel it is much clearer to users of the rule.

docs/rules/no-type-literal.md
weirdpattern: correct me if I'm wrong, isn't type aliases a much broader concept? a type alias could be:
- an actual alias `type foo = string | string[];`
- a callback `type foo = () => 'hello';`
- a literal `type foo = {};`

I used type literals, because the rule only covers the last case...

anyway I'm ok changing the text, just want to make sure you are ok with this

JamesHenry: (Sorry not on computer right now, so can't do much digging)

We have the "allowUnion" and "allowIntersection" configuration in this rule which broadens it from just a literal, right? Plus should we not actually be including your callback example in this rule as well?

e.g. should we not be preferring `func2` here?

https://www.typescriptlang.org/play/index.html#src=type%20func%20%3D%20()%20%3D%3E%20true%0D%0A%0D%0Ainterface%20func2%20%7B%0D%0A%20%20%20%20()%3A%20true%3B%0D%0A%7D%0D%0A%0D%0Aconst%20test%3A%20func%20%3D%20()%20%3D%3E%20true%0D%0Aconst%20test2%3A%20func2%20%3D%20()%20%3D%3E%20true

weirdpattern: this rule is based on the tslint rule (which includes only the literals), but yeah, we can broaden the scope by including the callback too. I think we should consider renaming the rule, though...

thoughts?

weirdpattern: by the way, the allowUnion and allowIntersection enable compositions that include a literal type. e.g.
`type foo = string & {};`

now I'm not so sure if we still need them if we go with literals and callbacks. I think we can rename the rule to no-type-aliases, with two properties allowTypeLiterals and allowCallbacks

weirdpattern:
@JamesHenry I did a couple of modifications to the rule based on the discussion we had few days ago.
Let me know what you think, I believe I cover all possible scenarios, but who knows...

NeoReyad:
@weirdpattern Would it be possible to ignore mapped types (https://www.typescriptlang.org/docs/handbook/advanced-types.html#mapped-types)? Not sure if this has been discussed yet or implemented in the current rule.

weirdpattern:
@NeoReyad nope, I did not include mapped types... but it shouldn't be too complicated to implementing.
Now, I know @JamesHenry is waiting on this rule to be merged to do a release, so maybe we can do this in v2 of the rule? 

What do you guys think?

weirdpattern:
on second thought, I think I'm gonna work on this tonight... I just realize the code doesn't handle those scenarios properly...

weirdpattern:
Ok, I added `allowMappedTypes` to the supported options.
If there are no more comments, this is ready for review!

weirdpattern:
Build is failing due to typescript version (2.0.3), works fine with latest version (2.1.x) as mapped types were introduced in this version [release notes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-1.html)

JamesHenry:
I'm sorry @weirdpattern but we cannot support TS 2.1.x in the parser yet - it is blocked on a TypeScript bug.

See the upgrade issue for more details: https://github.com/eslint/typescript-eslint-parser/issues/128

You will either have to remove the mapped types again or we can wait until after that issue is resolved to move forward with this PR. I'll leave it up to you!

NeoReyad:
@weirdpattern I really appreciate the work you done on this plugin and others. Thanks for taking the time to include mapped types. I have looked into the typescript JsDoc bug that is blocking the parser from upgrading. I will try and get a pull request submitted this weekend. Hopefully we can have TypeScript patched and released in a few weeks. 

I think the only part that is breaking is the tests and not the plugin itself since it does not depend on SyntaxKind only the node.type string. Maybe we could just make those tests conditional? 

If you decided to remove the feature that is fine too but it'd be cool if we could include it in a separate typescript 2.1 branch, like we currently do with the parser. This way I could use the mapped type rule today instead of forking the project in my own repo.

Thanks again.

weirdpattern:
@NeoReyad it looks like it's going to be a while before we can move to v2.1.x. This [bug](https://github.com/Microsoft/TypeScript/issues/13519) was labeled TypeScript 2.2 and there is no due date yet... 

Re creating a 2.1 branch for this rule (and any other requiring v2.1.x), I'm all for it, but I'm just a contributor here, so I cannot create branches on this repo. I leave the decision to @nzakas and/or @JamesHenry, for now I'm just going to leave this PR as is.

NeoReyad:
I see they have marked it as 2.2 milestone which i think is scheduled for march. One of the developer comments mentioned a quick fix. I will see if they can apply it to the 2.1 branch.

weirdpattern:
I guess I can remove the mapped types from this PR, then create a new one that includes everything. The latter can be merged once TS 2.2 gets released and the parser supports it.

I will work on this tomorrow.

weirdpattern:
@JamesHenry I removed mapped types from this PR, we can merge now.
I will submit a new PR in a second and am going to mark it as 2.1

JamesHenry:
Hey @weirdpattern, I am sorry for the insane delay on this. I have just merged in some important updates to the plugin from @corbinu, and released as 0.3.0.

Are you available to update this PR to include all the latest work?

weirdpattern:
@JamesHenry and @corbinu I've added the missing mapped types (see Feb 3 comment few lines above) and rebase the code to include the latest... we should be good to go, let me know otherwise...

JamesHenry:
Thanks, @weirdpattern!

PR #20, opened 2017-01-12
fix: no-angle-bracket-type-assertion fails with non-reference types (fixes #19)

### LINKED ISSUE #19: no-angle-bracket-type-assertion fails when typeAnnotation is not a reference type.
the following code throws an exception when no-angle-bracket-type-assertion is enabled

```ts
const a : number = <number>5;
```

but it works with reference types

```ts
interface A {}
const a : A = <A>{};
```

Already have a PR for this

weirdpattern:
Modify to no-angle-bracket-type-assertion rule
Add more tests for the rle

### DIFF
--- a/lib/rules/no-angle-bracket-type-assertion.js
+++ b/lib/rules/no-angle-bracket-type-assertion.js
@@ -19,17 +19,18 @@ module.exports = {
 
     create: function(context) {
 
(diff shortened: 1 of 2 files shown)


JamesHenry:
Thanks, @weirdpattern!

Issue #25, opened 2017-02-27
Rules missing in current NPM release


luchsamapparat:
Is there a timeline when we can expect the rules documented in the README.md to be released via NPM? Currently, when I install via NPM I only get explicit-member-accessibility and type-annotation-spacing.

soda0289:
@JamesHenry Could we get a release

JamesHenry:
I do not have access to that, and I'm not sure how @nzakas has been handling the releases given it is under his personal account. I will follow up with him.

Until then, npm installing from github should work for you @luchsamapparat

soda0289:
@JamesHenry Sorry to bother you then, will wait for @nzakas.

weirdpattern:
I think it will be a good idea to update the typescript parser version before we release.
Also, there are a couple of rules that still need approval :)

flying-sheep:
hi, any news? i’d like to see the `no-unused-vars` thing in a release

mattdell:
Same here. I'd like to use `no-unused-vars` but work only allows us to proxy to npmjs.org.

Need a release!

kayneb:
Likewise!!! Wanting to use `no-unused-vars` too.

soda0289:
Will try and get a new release this week. The ESLint team is working on creating an automated process for future releases.

soda0289:
Unfortunately it is not possible for the ESLint team to push a release to npm. @nzakas is currently taking a break from this project for health reasons and there might be a further delay until another release. 

In the mean time work can still continue work here and maybe create a fork that we can push to npmjs or release a tarball of the current master branch.

corbinu:
@soda0289 I am happy to do maintain a release fork @corbinu/eslint-plugin-typescript until nzakas is feeling better if that is cool with you all?

soda0289:
That would be fine with me.

@JamesHenry Would this be an acceptable solution, should we update the README here to indicate the current situation?

corbinu:
published as version 0.2.0

corbinu:
@soda0289 whats your npm username is it the same? I will add you with publish rights also.

JamesHenry:
We got all the permissions straightened out, I have just published `eslint-plugin-typescript@0.3.0`

PR #30, opened 2017-06-27
Update readme and package info to reflect the package name change

corbinu:
Renames the package to a eslint-plugin-ts so we can release new versions again. Solves #25 

@soda0289 you have been added to the npm repo so you can do releases also.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
-# eslint-plugin-typescript
+# eslint-plugin-ts/eslint-plugin-typescript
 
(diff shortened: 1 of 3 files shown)


soda0289:
Thanks! Will try and spend some time here after I get caught up with the issues in the parser.

JamesHenry:
Thanks for driving this @corbinu!

I have chatted with Nicholas and he has now given me access to publish this package on NPM, so we can continue with the same name and just get everything fixed up as per your other PR.

corbinu:
ok yay! I hope he is doing ok we just didn't want to bother him!

PR #31, opened 2017-06-27
Chore: Upgrade deps to eslint 4 and typescript-parser 4

corbinu:
This should land after #30 I updated the deps. Cleaned up the linting to fit eslint-config-eslint 4 and then fixed the broken tests. I did test against both eslint 3 and 4. We should be ready to add new rules now!

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -1,4 +1,7 @@
-extends: eslint
-env:
-  es6: true
-  node: true
+root: true
+
+plugins:
+    - node
+extends:
+    - "eslint"
+    - "plugin:node/recommended"
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+package-lock = false
--- a/docs/rules/type-annotation-spacing.md
+++ b/docs/rules/type-annotation-spacing.md
@@ -3,7 +3,7 @@
 Type annotations in TypeScript tend to have a very specific format, where the colon appears directly after the identifier or function, followed by a space, followed by the type. For example:
 
 ```ts
-var foo: string = "bar";
+let foo: string = "bar";
 
 function foo(a: string): string {
     // code
@@ -19,7 +19,7 @@ This rule aims to enforce specific spacing patterns around type annotations. As 
 The following patterns are considered warnings:
 
 ```ts
-var foo:string = "bar";
+let foo:string = "bar";
 
 function foo(a : string):string {
     // code
@@ -29,7 +29,7 @@ function foo(a : string):string {
 The following patterns are not warnings:
 
 ```js
-var foo: string = "bar";
+let foo: string = "bar";
 
 function foo(a: string): string {
     // code
--- a/lib/index.js
+++ b/lib/index.js
@@ -8,8 +8,8 @@
 // Requirements
 //------------------------------------------------------------------------------
 
-var requireIndex = require("requireindex"),
-    path = require("path");
+const requireIndex = require("requireindex");
+const path = require("path");
 
 //------------------------------------------------------------------------------
 // Plugin Definition
--- a/lib/rules/explicit-member-accessibility.js
+++ b/lib/rules/explicit-member-accessibility.js
@@ -17,7 +17,7 @@ module.exports = {
         schema: []
     },
 
-    create: function(context) {
+    create(context) {
 
         //----------------------------------------------------------------------
         // Helpers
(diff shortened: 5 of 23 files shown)


corbinu:
Spoke to soon, but looks like it is actually the test harness @JamesHenry do you have any idea what might have broken it?

soda0289:
I can take a look. A lot of re-factoring was done to the Linter class in eslint don't think too much changed with RuleTester.

JamesHenry:
Thanks a lot @corbinu! As mentioned on the other PR, we will now not need to change the name, so please remove the relevant changes from the PR whenever you get chance to fix up the ruleTester based on @soda0289 suggestion

corbinu:
@JamesHenry @soda0289 Thanks I just rolled it back let me figure out what broke the no-unused and then will squash it down and we are good to go

corbinu:
@soda0289 That change seems to be causing only the main `no-unused-vars` to run as when I try to log out in `typescript/no-unused-vars` to see why the decorators aren't getting marked I never get any output

soda0289:
@corbinu I'm not sure what the test was supposed to be checking for. I guess it wants both rules to be run at the same time.

This line:
`ruleTester.run("typescript/no-unused-vars", ruleNoUnusedVars,`

Is loading the original rule since: ruleNoUnusedVars is the original no-unused-vars rule. Maybe we should change that part as well. Can spend some more time tonight digging into it,

corbinu:
@soda0289 Ok I think I know whats happening the change is causing eslint to overwrite typescript/no-unused-vars as the original rule. We need to change it back to no-unused-vars and then figure out why eslint isn't finding that rule even though it is defined on line 16 with this

  `eslint.defineRule("typescript/no-unused-vars", rule);`

soda0289:
@corbinu I think what might be happening is that RuleTester is creating its own Linter instance and eslint has a different Linter instance. In eslint v4 Linter is no longer a singleton meaning multiple instances can exists.

Instead of doing `eslint.defineRule()` try `ruleTester.linter.defineRule()`

corbinu:
@soda0289 Haha just did that and was about to comment I found it. Thanks for all your help!

corbinu:
@soda0289 @JamesHenry All set and ready for release! I bumped the version to 0.3.0 since we did release a 0.2.0 under my scope before.

corbinu:
Also fixed bugs against both parser master and ts-2.4

soda0289:
Theres a breaking change in the parser that will cause the type-annotation spacing rule to break. Not sure if we should fix that now or after we release a new version of the parser.

https://github.com/eslint/typescript-eslint-parser/pull/319

corbinu:
@soda0289 I don't think there is a way to do it without breaking support for version 3 of the parser so lets merge this and then we can release a version 0.4.0 that supports version 4 of the parser and typescript 2.4

corbinu:
@JamesHenry Now that the breaking change is in parser 4.0.0 I do think I should fix that first.

Also I am really sad to hear that but will do for parity.

corbinu:
@JamesHenry Wouldn't it make more sense to add package-lock.json to .gitignore that way at least reinstalls during development would be faster?

soda0289:
@corbinu Based on this comment: https://github.com/eslint/eslint/issues/8742#issuecomment-308927735
event if it is added to .gitignore it will still be included in the commit when `npm version` is called.

corbinu:
@soda0289 Yes but my understanding is that specifically has to do with eslint's release system which we aren't using here. Also package-lock for dependencies is not respected on npm install so it ending up in the release or not does absolutely nothing. It only affects when somebody git clones the repo.

So what we are talking about is A) if it is committed then when any of us forks this repo we are all working of the same dep tree B) it is not committed and at least we each get a faster npm install on subsequent installs in our own versions. 

Also this comment 
```
I think we should not ignore `package-lock.json`. If `package-lock.json` exists, npm seems to ignore `package.json`. As a result, I guess local `package-lock.json` will cause confusing if `git pull` changes `package.json`.
```

Is no longer true.

soda0289:
@corbinu I see your point. I'm not too familiar on the details of npm and package-lock.json just thought I'd share the reason why eslint did not use .gitignore

corbinu:
@soda0289 I just wish I had seen that issue before it was merged would have brought all this up there. However if you guys want it to match I understand

corbinu:
@JamesHenry All set for release but I still don't like it ;)

JamesHenry:
Thanks! Published as `eslint-plugin-typescript@0.3.0`

PR #34, opened 2017-08-06
no-use-before-define

j-f1:
This adds the `no-use-before-define` rule from core ESLint, with a new `typedefs` option to ignore errors coming from `type` declarations.


weirdpattern:
Hi @j-f1, there are no tests in your PR. 
Do you mind adding a couple, including, but not limited to, the new typedefs option.

j-f1:
@weirdpattern How does that look?

corbinu:
LGTM

JamesHenry:
Thanks a lot for contributing, @j-f1!

As I mentioned on another PR just now, I don't want to slow down this awesome momentum so I am going to go ahead and merge this without the most detailed of reviews 😄

Even though the plugin has been around for a while, it is still early days, so I would rather just get this in and we can work on problems, if they arise.

JamesHenry:
I just published version 0.4.0 to npm and this PR is included in it 👍

PR #35, opened 2017-08-08
Fix: no-explicit-any Rule does not check generic type (fix #26)

### LINKED ISSUE #26: no-explicit-any Rule does not check generic type
For example:

```ts
function test(): Promise<any> {
return new Promise();
}
```

Maybe we can include an extra option?

weirdpattern:
This change improves the existing no-explicit-any rule by adding:

- Check for TSUnionTypes
- Check for TSIntersectionTypes
- Check for TypeReferences (Array<*>) and TSArrayType (any[])
- Check for Arguments in class constructors 
- Check for Properties in classes, interfaces and type aliases

This also adds more tests cases

### DIFF
--- a/docs/rules/no-explicit-any.md
+++ b/docs/rules/no-explicit-any.md
@@ -14,21 +14,77 @@ an `any` type from being implied by the compiler, but doesn't prevent
 The following patterns are considered warnings:
 
 ```ts
-const age : any = "seventeen"
+const age: any = "seventeen"
 ```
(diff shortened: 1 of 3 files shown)


corbinu:
LGTM

JamesHenry:
Thanks!

JamesHenry:
I just published version 0.4.0 to npm and this PR is included in it 👍

Issue #42, opened 2017-08-13
type-annotation-spacing fails for "as T" syntax


nevir:
`type-annotation-spacing` doesn't play nice with `as T` type assertions:

```ts
const foo = {} as Foo;
//            ^ Unexpected space before the colon.
```

weirdpattern:
This seems like another of those edge cases for the TypeAnnotation handler.
I will review this tomorrow and include it in the overrides that I just added in PR #43

nevir:
Thanks!

weirdpattern:
hey @JamesHenry you can close this bug too, it was merged in #43

PR #43, opened 2017-08-13
Update: Add override option to type-annotation-spacing (fixes #41 and #42)

### LINKED ISSUE #41: type-annotation-spacing doesn’t work for function typedefs
This throws an error when it shouldn’t:

```ts
type Foo = (bar: string) => string
//                      ^
// Unexpected space before the colon (typescript/type-annotation-spacing) 
```

Maybe there should be a separate set of options for the `:` annotation and the `=>` annotation, with the default being a space after but not before the `:` and a space before and after the `=>`.

weirdpattern:
Add an override option to the type-annotation-spacing rule
The option allows the developer to override the behavior of type annotations and function types

e.g.
type annotation: const a : number 
function type: type a = (s: string) => string;


corbinu:
LGTM

PR #46, opened 2017-08-13
Fix: fixes bound methods ordering (fixes #45)

### LINKED ISSUE #45: member-ordering fails for bound methods
With the default configuration, this fails:

```ts
class Foo {
  constructor () {}
  method () {}
  boundMethod = () => {}
}
```

It claims that the property `boundMethod` should be placed before the constructor. To fix this, the code in `getNodeType` should check if the value of the property is a `FunctionExpression` or `ArrowFunctionExpression`, then assume that it is a method.

weirdpattern:
Fixes an issue where bound methods were processed as properties in classes, interfaces and type literals.

### DIFF
--- a/lib/rules/member-ordering.js
+++ b/lib/rules/member-ordering.js
@@ -111,6 +111,7 @@ module.exports = {
 
         const options = context.options[0] || {};
 
+        const functionExpressions = ["FunctionExpression", "ArrowFunctionExpression"];
         const defaultOrder = [
             "public-static-field",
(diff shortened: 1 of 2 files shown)


weirdpattern:
@j-f1 you are absolutely right... fixed

weirdpattern:
Yeah, the white spaces were removed by the .editorconfig, we will probably see a lot of these, especially when reviewing old files...

PR #47, opened 2017-08-13
Chore: Add prettier and reformat existing code

aboyton:
Added prettier to be part of linting. Automatically fixed most and made a couple of minor changes to remove some `no-multi-assign` in `lib/rules/no-use-before-define.js` that prettier doesn't format beautifully.


weirdpattern:
Can we include the tests as well? 
The current package.json lints both lib/ and test/ folders.

aboyton:
Ah, there was a bug in `lint:fix` in `package.json` that meant that they were skipped. Fixed now. I was wondering why this didn't work.

Not much of a contribution yet, just testing the waters :)

Is there a gitter channel for this plugin where I can ask questions, or should I post issues?

weirdpattern:
Nah, everything is a contribution... welcome to the team.
I don't think we have gitter channel, at least not that I'm aware of. @JamesHenry do we?
I guess you can create a bug with your questions... ☺️

.eslintrc.yml
```diff
@@ -2,6 +2,15 @@ root: true
...
+    - prettier
 extends:
     - "eslint"
     - "plugin:node/recommended"
+    - "prettier"
+rules:
+    prettier/prettier:
+    - error
+    - tabWidth: 4
+      printWidth: 120
```
JamesHenry: Hmm please could explain this choice? The default of 80 in prettier was chosen for a reason, and increasing it by 50% seems unnecessary

JamesHenry:
@aboyton Thanks for contributing! There is not currently a separate forum for this plugin, so yeah feel free to open issues with questions/blockers

.eslintrc.yml
aboyton: I'm very happy to make it 80. The only reason I made it 120 was that I was trying to conform to what the code itself _seemed_ to be doing and making it 80 made the diff bigger. I'll change it and rebase the diff.

While we're at it, any reason to increase the default tab width by 100% 😉

package.json
```diff
@@ -12,7 +12,7 @@
   "main": "lib/index.js",
   "scripts": {
     "lint": "eslint lib/ tests/",
-    "lint:fix": "eslint lib/ test/ --fix",
+    "lint:fix": "eslint lib/ tests/ --fix",
```
aboyton: This was a bug.

.eslintrc.yml
weirdpattern: @aboyton check the last point in style ☺️
[TypeScript guidelines](https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines)

aboyton:
Thanks for the comments. I've switched it over to the default 80 char width. Should we go for the default tab width too?

weirdpattern:
Pretty sure you saw this, but it's kind of buried in the review comments, so I'm reposting it..
@aboyton check out the last point in style [TypeScript guidelines](https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines)

weirdpattern:
Deleted my previous comment as it doesn't make any sense LOL... those are TS guidelines not JS which is what we are using... sorry been working with TS for a long time... lines are blurry now...

JamesHenry:
Thanks for the great work @aboyton, I'd like to get this in ASAP, so nobody works on anything in parallel to this, so the tab width could potentially be addressed in the follow up PR around enforcing prettier use on staged files: https://github.com/nzakas/eslint-plugin-typescript/pull/52

PR #55, opened 2017-08-19
Chore: Bump deps and fix errors in no-explicit-any

### LINKED ISSUE #54: no-explicit-any Broken in typescript paser 5.0.0 and greater
Just a note looks like no-explicit-any has been broken by typescript paser 5.0.0

corbinu:
Bumps the parser and other deps and fixes errors in no-explicit-any. Closes #54

### DIFF
--- a/lib/rules/no-explicit-any.js
+++ b/lib/rules/no-explicit-any.js
@@ -42,10 +42,14 @@ module.exports = {
                 node.types.forEach(type => {
                     checkGenericNodeForAnnotation(type);
                 });
-            } else if (node.type === "TSTypeReference" && node.typeParameters) {
-                node.typeParameters.params.forEach(param => {
-                    checkGenericNodeForAnnotation(param);
-                });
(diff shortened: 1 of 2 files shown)


weirdpattern:
that was probably my fault, I copy-pasted the code then manually formatted it... sorry!

j-f1:
It might be a good idea to enable [Greenkeeper](https://GitHub.com/apps/greenkeeper) to avoid having to do this in the future. It’ll automatically run your tests with the latest version of each dependency, and open an issue if an in-range dependency breaks the build. If a dependency updates out of the range specified in your package.json, you’ll get a PR to update it automatically.

JamesHenry:
@j-f1 Greenkeeper is great but unfortunately I do not have admin access to the repo, so cannot enable such things.

j-f1:
cc @nzakas

> Greenkeeper is great but unfortunately I do not have admin access to the repo, so cannot enable such things.

PR #64, opened 2017-08-25
Fix: fixes incorrectly marked function call as misplaced (fixes #63)

### LINKED ISSUE #63: Rule adjacent-overload-signatures incorrectly marks a function call as misplaced
The code must validate the name returned by the `getMemberName` function before storing it in the seen variables array. Nulls are not variables... 

Error Message:
  All 'null' signatures should be adjacent.

weirdpattern:
Missed a validation that was required to prevent null values to be stored in the seen array


weirdpattern:
And another one for export statements (re-exports)

PR #68, opened 2017-09-10
New: add no-array-constructor rule

j-f1:
Copied from ESLint core, except it allows `new Array<Foo>()` annotations.

### DIFF
--- a/docs/rules/no-array-constructor.md
+++ b/docs/rules/no-array-constructor.md
@@ -1,0 +1,34 @@
+# disallow generic `Array` constructors (no-array-constructor)
+
+Use of the `Array` constructor to construct a new array is generally discouraged in favor of array literal notation because of the single-argument pitfall and because the `Array` global may be redefined. Two exceptions are when the Array constructor is used to intentionally create sparse arrays of a specified size by giving the constructor a single numeric argument, or when using TypeScript  type parameters to specify the type of the items of the array (`new Array<Foo>()`).
+
+## Rule Details
+
+This rule disallows `Array` constructors.
+
+Examples of **incorrect** code for this rule:
+
+```ts
+/*eslint no-array-constructor: "error"*/
+
+Array(0, 1, 2)
+new Array(0, 1, 2)
+```
+
+Examples of **correct** code for this rule:
+
+```ts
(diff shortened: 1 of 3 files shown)


lib/rules/no-array-constructor.js
```diff
@@ -0,0 +1,62 @@
...
+
+    create(context) {
+        /**
+         * Disallow construction of dense arrays using the Array constructor
+         * @param {ASTNode} node node to evaluate
+         * @returns {void}
+         * @private
+         */
+        function check(node) {
+            if (
```
weirdpattern: This is probably a long shot, but I'm going to say it anyway... 

The only valid constructors should be `new Array(number)` and `new Array(identifier)`, but the code allows `new Array('hello')`, `new Array(true)`, etc...

I know we don't have type information available at runtime to validate identifiers (e.g. `new Array(identifier)`, nothing we can do about them without complicating things), but I still think we can validate literals 😃, perhaps a typeof node.arguments[0].value === "number" when node.arguments[0].type === "Literal"?

Thoughts?

docs/rules/no-array-constructor.md
```diff
@@ -0,0 +1,34 @@
+# disallow generic `Array` constructors (no-array-constructor)
```
weirdpattern: Can we capitalize this?

lib/rules/no-array-constructor.js
j-f1: It seems like doing `new Array(<anything that’s not a number>)` will just return an array with that parameter as the first element.

weirdpattern: That is correct. But isn't creating empty arrays with the given length what we are trying to validate? 
> Two exceptions are when the Array constructor is used to intentionally create sparse arrays of a specified size by giving the constructor a single numeric argument

j-f1: I’d open this as an issue on eslint/eslint.

JamesHenry:
Really sorry for the delay, I buried myself pretty badly for the last month and open-source had to take a back seat. Light at the end of the tunnel now! P.S. Apologies that you will see this comment multiple times 😅

PR #71, opened 2017-09-13
Chore: Update dependencies

weirdpattern:
Updates dependencies: eslint, parser, typescript and others


j-f1:
Seems like this is the kind of thing to use [Greenkeeper](https://greenkeeper.io) for.

corbinu:
eslint policy is not to use greenkeeper https://github.com/eslint/eslint/issues/8685

PR #75, opened 2017-09-27
Fix no-triple-slash-reference

mightyiam:
Trying this rule, I had to fix it by replacing `getComments` with `getCommentsBefore`.

Anyway, [`getComments` seems deprecated](https://eslint.org/docs/developer-guide/working-with-rules#deprecated).

### DIFF
--- a/tests/lib/rules/no-triple-slash-reference.js
+++ b/tests/lib/rules/no-triple-slash-reference.js
@@ -30,6 +30,17 @@ ruleTester.run("no-triple-slash-reference", rule, {
     ],
     invalid: [
         {
+            code: '/// <reference path="Animal" />',
+            parser: "typescript-eslint-parser",
+            errors: [
+                {
+                    message: "Do not use a triple slash reference",
+                    line: 1,
+                    column: 1
+                }
+            ]
+        },
+        {
             code: '/// <reference path="Animal" />\nlet a',
             parser: "typescript-eslint-parser",
             errors: [
--- a/tests/lib/rules/no-triple-slash-reference.js
+++ b/tests/lib/rules/no-triple-slash-reference.js
@@ -30,7 +30,7 @@ ruleTester.run("no-triple-slash-reference", rule, {
     ],
     invalid: [
         {
-            code: '/// <reference path="Animal" />',
+            code: '/// <reference path="Animal" />\nlet a',
             parser: "typescript-eslint-parser",
             errors: [
                 {
--- a/tests/lib/rules/no-triple-slash-reference.js
+++ b/tests/lib/rules/no-triple-slash-reference.js
@@ -41,12 +41,15 @@ ruleTester.run("no-triple-slash-reference", rule, {
             ]
         },
         {
-            code: '/// <reference path="Animal" />\nlet a',
+            code: `
+/// <reference path="Animal" />
+let a
+            `,
             parser: "typescript-eslint-parser",
             errors: [
                 {
                     message: "Do not use a triple slash reference",
-                    line: 1,
+                    line: 2,
                     column: 1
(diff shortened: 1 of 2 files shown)


tests/lib/rules/no-triple-slash-reference.js
```diff
@@ -30,7 +30,7 @@ ruleTester.run("no-triple-slash-reference", rule, {
     ],
     invalid: [
         {
-            code: '/// <reference path="Animal" />',
+            code: '/// <reference path="Animal" />\nlet a',
```
mightyiam: This fails with `commentsBefore`. Can guess why.

tests/lib/rules/no-triple-slash-reference.js
```diff
@@ -39,6 +39,17 @@ ruleTester.run("no-triple-slash-reference", rule, {
                     column: 1
                 }
             ]
+        },
+        {
+            code: '/// <reference path="Animal" />\nlet a',
```
mightyiam: This fails with `getComments(program).leading` because `leading` seems empty.

weirdpattern: Can we use a template literal instead of the \n?

```javascript
code: `
/// <reference path="Animal" />
let a;
`
```

I think it makes it more readable.

weirdpattern:
Small change after which this should be OK, thanks for the contribution @mightyiam!!

tests/lib/rules/no-triple-slash-reference.js
mightyiam: Sure. Tomorrow.

JamesHenry:
Thanks a lot for contributing, @mightyiam!

Seems like you are planning on making one more change, so I won't merge just yet.

tests/lib/rules/no-triple-slash-reference.js
mightyiam: Actually, this seems to result in a different string, because the leading whitespaces and the extra newlines are not trimmed. There are template tags like [this one](https://www.npmjs.com/package/common-tags#stripindents) but seems overkill to me.

weirdpattern: You can move the code all the way to the left, check how we do it in other tests, most of them use this approach

weirdpattern: I agree adding the template would be an overkill, that's why we just ignore the leading spaces by removing the indentation

weirdpattern: By the way, I created bug #81 to standardize the tests using this same approach and some other suggestions made by @j-f1... Currently working on this...

weirdpattern:
Awesome, thanks for taking care of this.
LGTM

tests/lib/rules/no-triple-slash-reference.js
```diff
@@ -39,6 +39,20 @@ ruleTester.run("no-triple-slash-reference", rule, {
                     column: 1
                 }
             ]
+        },
+        {
+            code: `
+/// <reference path="Animal" />
+let a
+            `,
```
mightyiam: Fixed the code style issue here.

mightyiam: Had to change the error line from `1` to `2`, for this.

weirdpattern:
LGTM

j-f1:
@mightyiam Can you close and reopen this PR to restart the CI builds? They should be passing now.

mightyiam:
@j-f1 done. Passing.

JamesHenry:
Thanks so much for your patience on this, @mightyiam!

PR #80, opened 2017-10-06
Fix: Handles type spacing on TSParenthesizedType expressions (fixes #79)

### LINKED ISSUE #79: 0.8.0: type-annotation-spacing bug
In 0.7.0, this line would pass lint checking:

```javascript
  resolver: (() => PromiseLike<T>) | PromiseLike<T>;
```

In 0.8.0, I get this error:
>   13:13  error  Expected a space after the '('  typescript/type-annotation-spacing

where `(` in the error represents the first parenthesis after the `: `

weirdpattern:
Handles constructors like.

```ts
interface A {
    resolver: (() => PromiseLike<T>) | PromiseLike<T>;
}
```

By excluding punctuators other than `:` and `=>` from the validation

### DIFF
--- a/tests/lib/rules/type-annotation-spacing.js
+++ b/tests/lib/rules/type-annotation-spacing.js
@@ -20,6 +20,14 @@ const ruleTester = new RuleTester();
 ruleTester.run("type-annotation-spacing", rule, {
     valid: [
         {
+            code: `
+interface resolve {
+    resolver: (() => PromiseLike<T>) | PromiseLike<T>;
+}
+            `,
+            parser: "typescript-eslint-parser"
+        },
+        {
             code: "const foo = {} as Foo;",
             parser: "typescript-eslint-parser"
         },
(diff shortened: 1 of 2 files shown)


tests/lib/rules/type-annotation-spacing.js
```diff
@@ -20,6 +20,14 @@ const ruleTester = new RuleTester();
 ruleTester.run("type-annotation-spacing", rule, {
     valid: [
         {
+            code: `
+interface resolve {
+    resolver: (() => PromiseLike<T>) | PromiseLike<T>;
+}
+            `,
+            parser: "typescript-eslint-parser"
```
j-f1: If you pass `{ parser: "typescript-eslint-parser" }` to the `RuleTester` constructor, it will include that option in all tests, which will let you just have an array of code blocks to test.

weirdpattern: Yeah, I saw this in one of your tests and though of doing the same for new rules.
We could probably create a new bug to refactor the tests, right now we have different styles and is driving me crazy... 

are you OK with leaving this as is for now? I will work on the refactor during the weekend

j-f1: > are you OK with leaving this as is for now?

Yes 👍

JamesHenry:
Thanks a lot for your patience on this, @weirdpattern!

Issue #82, opened 2017-10-08
no-unused-vars fails to mark interfaces as used when destructuring in eg. a function parameter


lostfictions:
With these versions:
```
"dependencies": {
  "eslint": "^4.8.0",
  "eslint-plugin-typescript": "^0.8.0",
  "typescript-eslint-parser": "^8.0.0",
  "typescript": "^2.5.3"
}
```

and these rules:
```
"rules": {
    "no-unused-vars": "warn",
    "typescript/no-unused-vars": "warn"
}
```

and this test case:

`test.ts`
```typescript
import { ReproInterface } from './test2'

const x = ({ a } : ReproInterface) => a
```

`test2.ts`
```typescript
export interface ReproInterface {
  a : number
}
```

In `test.ts`, `ReproInterface` is incorrectly marked as unused. If the function parameter is not destructured, the warning goes away.

weirdpattern:
Hi @lostfictions, thank you for reporting this. I will take a look at this some time tomorrow...

PR #84, opened 2017-10-08
Destructured param type annotation not marked used

mightyiam:
Here is a failing seemingly valid test code.

### DIFF
--- a/tests/lib/rules/no-unused-vars.js
+++ b/tests/lib/rules/no-unused-vars.js
@@ -383,6 +383,14 @@ ruleTester.run("no-unused-vars", ruleNoUnusedVars, {
         },
         {
             code: [
+                "import { ReproInterface } from 'ReproInterface'",
+                "const x = ({ a = null } : { a: ReproInterface }) => a",
+                "console.log(x)"
+            ].join("\n"),
+            parser
+        },
+        {
+            code: [
                 "import { Nullable } from 'nullable'",
                 "import { SomeOther } from 'some'",
                 "import { Another } from 'some'",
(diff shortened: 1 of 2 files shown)


tests/lib/rules/no-unused-vars.js
```diff
@@ -376,6 +376,14 @@ ruleTester.run("no-unused-vars", ruleNoUnusedVars, {
         {
             code: [
                 "import { Nullable } from 'nullable'",
+                "const foo = ({ nullable }: Nullable)  => nullable",
```
j-f1: `Nullable) => Nullable`? (remove the space; capitalize `nullable`)

mightyiam: That won't be what I meant.

`const foo = ` is assignment. `(` starts arrow function. `{ nullable }` is destructuring assignment. `: Nullable` is type annotation for the value before destructuring. The rest is clear. Although I do have an extra space before the arrow (negligible?).

weirdpattern: @j-f1 I think you are looking at this as a declaration, whilst @mightyiam defined as an actual arrow function... there is one more thing I noticed though.. the last call foo(); won't compile in typescript as an argument is required by the arrow function...

weirdpattern: Also, the extra whitespace won't really matter, but if you think about this... we are developing a plugin for a linting tool... it would by kind of ironic not to have our own code linted... what do you think?

mightyiam:
@weirdpattern and @j-f1 I have fixed the issues you brought up. A space and the function call.

Your turn to fix the bug :laughing:.

weirdpattern:
Thank you @mightyiam I will take a look at this tomorrow

weirdpattern:
Ok, I submitted a PR in @mightyiam to update his branch, and finally this PR.
This should take care of #82

PR #88, opened 2017-10-20
Fix: Mark qualified name type annotations as used

timothykang:


timothykang:
Good call, updated.

mightyiam:
Does this fix any filed issue?

timothykang:
None that I could find.

I encountered the bug myself and was about to create an issue and PR with a failing test, but then thought why not just fix it while I was at it.

JamesHenry:
Thanks a lot for your patience on this, @timothykang!

PR #91, opened 2017-11-05
failing test: type-annotation-spacing: keyof considered like arrow

mightyiam:
Perhaps there should be more override types to go with `colon` and `arrow`?

This particular example is straight out of no-one's style guide, I assume.


mightyiam:
This used to fail. I guess type-annotation-spacing got some fixes recently. Now this is just another valid test case.

mightyiam:
Take it or leave it.

PR #93, opened 2017-11-20
prettier

mightyiam:
Closes #94.


mightyiam:
In case we also would like to pin the prettier version so that this does't happen, there is #94.

PR #98, opened 2017-12-01
Fix class-name-casing cannot recognize PascalCase containing digits

dobogo:
When `class-name-casing` is enabled, `class Foo2 {}` will be reported as invalid.


JamesHenry:
Thanks, @dobogo!

Issue #102, opened 2017-12-22
`typescript/explicit-member-accessibility` false positive in mixed TS & JS codebase


scottohara:
**What version of TypeScript are you using?**
2.6.2

**What version of typescript-eslint-parser are you using?**
9.0.0

**What version of eslint-plugin-typescript are you using?**
0.8.1

**What code were you trying to parse?**

```javascript
/* foo.js */
export default class Foo {
  constructor() { ... }
}
```

```javascript
/* bar.ts */
export default class Bar {
  public constructor() { ... }
}
```

```
eslint --rule 'typescript/explicit-member-accessibility: 2' '*.{ts,js}'
```

**What did you expect to happen?**
No errors/warnings

**What happened?**

```
Missing accessibility modifier on method definition constructor
```

In a codebase that includes both JS and TS files, enabling the `typescript/explicit-member-accessibility` rule triggers false positives on all classes in JS files, because the class members don't have accessibility modifiers (e.g. `public`, `private`, `protected`); despite the fact that such modifiers are invalid syntax in Javascript.

scottohara:
Could this be addressed by simply checking that the source file name ends with `*.ts`, or would just checking the file extension be too naive?

e.g.
```javascript
function checkMethodAccessibilityModifier(methodDefinition) {
  if (!methodDefinition.accessibility && /\.ts$/test(context.getFilename()) {
    ...
  }
}
```

If so, I'd be happy to submit a PR to this effect.

PR #103, opened 2017-12-24
New: add explicit-function-return-type rule

scottohara:
Adds a new rule that ensures that all function declarations, function expressions, arrow functions, class methods etc. include a type annotation for the return value (including `void` for functions that don't return a value).

The following patterns are considered warnings:
```ts
// Should indicate that no value is returned (void)
function test() {
    return;
}

// Should indicate that a number is returned
var fn = function() {
    return 1;
}

// Should indicate that a string is returned
var arrowFn = () => 'test';

class Test {
    // Should indicate that no value is returned (void)
    method() {
        return;
    }
}
```

The following patterns are not warnings:

```ts
// No return value should be expected (void)
function test(): void {
    return;
}

// A return value of type number
var fn = function(): number {
    return 1;
}

// A return value of type string
var arrowFn = (): string => 'test';

class Test {
    // No return value should be expected (void)
    method(): void {
        return;
    }
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,7 @@ Then configure the rules you want to use under the rules section.
 <!-- begin rule list -->
 * [`typescript/adjacent-overload-signatures`](./docs/rules/adjacent-overload-signatures.md) — Require that member overloads be consecutive
 * [`typescript/class-name-casing`](./docs/rules/class-name-casing.md) — Require PascalCased class and interface names (`class-name` from TSLint)
+* [`typescript/explicit-function-return-type`](./docs/rules/explicit-function-return-type.md) — Require explicit return types on functions and class methods
 * [`typescript/explicit-member-accessibility`](./docs/rules/explicit-member-accessibility.md) — Require explicit accessibility modifiers on class properties and methods (`member-access` from TSLint)
 * [`typescript/interface-name-prefix`](./docs/rules/interface-name-prefix.md) — Require that interface names be prefixed with `I` (`interface-name` from TSLint)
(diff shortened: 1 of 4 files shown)


ajafff:
Isn't this already covered by the `typedef` rule?

scottohara:
In the initial commit, I missed the fact that class constructors and property setters cannot have a return type annotation in Typescript;  so these two specific types of `FunctionExpression` nodes are now ignored by the rule.

e.g.
```ts
// Valid
class Test {
  constructor() {...}  // OK...constructors don't need a return type
  set foo() {...} // OK...property setters don't need a return type
  get foo(): number {...}  // OK...property getter has a return type
}

// Invalid
class Test {
  constructor() {...}  // OK...constructors don't need a return type
  set foo() {...} // OK...property setters don't need a return type
  get foo() {...} // Error...property getter has no return type
}
```

scottohara:
@ajafff - what is the `typedef` rule?

ajafff:
@scottohara TSLint has a rule called `typedef` that enforces explicit type annotations in the certain locations.
I just noticed after commenting that this PR is for the eslint plugin.

scottohara:
I should also point out that, like #102, this rule in its current form will trigger false positives on any `*.js` files (for the obvious reason that return type annotations are invalid JS syntax).

Whatever fix is agreed as appropriate for #102 (such as limiting the rule to only `*.ts` files, e.g. `/\.ts$/.test(context.getFilename())`) will equally need to be applied here.

j-f1:
It should also run on `.tsx` files.

et:
Can you release a new version that has this included?

JamesHenry:
@et just released `0.9.0`

et:
Awesome! I will give a go right now 😄

PR #104, opened 2017-12-27
Update: explicit-member-accessibility only check *.ts and *.tsx files (Fixes #102)

### LINKED ISSUE #102: `typescript/explicit-member-accessibility` false positive in mixed TS & JS codebase

scottohara:
Modifies the `explicit-member-accessibility` rule by restricting it to only checking *.ts and *.tsx files (Fixes #102).

Note: the proposed new rule `explicit-function-return-type` (#103) has similar functionality (both rules duplicate the `isTypescript(filename)` function.  It may be desirable to move this duplicated function into `utils.js`, so it is available to any rule that needs to limit itself to only operating against Typescript files.

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -1,3 +1,11 @@
 "use strict";
 
 exports.tslintRule = name => `\`${name}\` from TSLint`;
+
+/**
+ * Check if the context file name is *.ts or *.tsx
+ * @param {string} fileName The context file name
+ * @returns {boolean} `true` if the file name ends in *.ts or *.tsx
+ * @private
+ */
+exports.isTypescript = fileName => /\.(ts|tsx)$/.test(fileName);
--- a/lib/util.js
+++ b/lib/util.js
@@ -8,4 +8,4 @@ exports.tslintRule = name => `\`${name}\` from TSLint`;
  * @returns {boolean} `true` if the file name ends in *.ts or *.tsx
  * @private
  */
-exports.isTypescript = fileName => /\.(ts|tsx)$/.test(fileName);
+exports.isTypescript = fileName => /\.tsx?$/.test(fileName);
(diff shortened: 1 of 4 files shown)


JamesHenry:
Great point, please could you do that in this PR now that the other one has been merged?

scottohara:
Absolutely. Will get onto this when I return from vacation next week.

Thanks.

scottohara:
Duplicated `isTypescript()` function has been moved to `util.js`, and both the `explicit-function-return-type` and `explicit-member-accessibility` rules have been updated accordingly.

lib/util.js
```diff
@@ -1,3 +1,11 @@
...
 
 exports.tslintRule = name => `\`${name}\` from TSLint`;
+
+/**
+ * Check if the context file name is *.ts or *.tsx
+ * @param {string} fileName The context file name
+ * @returns {boolean} `true` if the file name ends in *.ts or *.tsx
+ * @private
+ */
+exports.isTypescript = fileName => /\.(ts|tsx)$/.test(fileName);
```
j-f1: How about `/\.tsx?$/`?

scottohara: Yep, that's better.  Thanks @j-f1.

Issue #111, opened 2018-03-06
Rule Idea: generic-type-naming


nevir:
Proposal:

---

# Enforces naming of generic type variables. (generic-type-naming)

It can be helpful to enforce a consistent naming style for generic type variables used within a type. For example, prefixing them with `T` and ensuring a somewhat descriptive name, or enforcing Hungarian notation.

## Rule Details

This rule allows you to enforce conventions over type variables. By default, it does nothing.

## Options

The rule takes a single string option, which is a regular expression that type variables should match.

Examples of **correct** code with a configuration of `'^T[A-Z][a-zA-Z]+$'`:

```ts
type ReadOnly<TType extends object> = {
  readonly [TKey in keyof TType]: TType[TKey];
}

interface SimpleMap<TValue> {
  [key: string]: TValue;
}
```

Examples of **incorrect** code with a configuration of `'^T[A-Z][a-zA-Z]+$'`:

```ts
type ReadOnly<T extends object> = {
  readonly [Key in keyof T]: TType[Key];
}

interface SimpleMap<T> {
  [key: string]: T;
}
```

## When Not To Use It

If you do not want to enforce a naming convention for type variables.

## Further Reading

* [TypeScript Generics](https://www.typescriptlang.org/docs/handbook/generics.html)

PR #113, opened 2018-03-09
New: Add generic-type-naming (#111)

unlight:
https://github.com/nzakas/eslint-plugin-typescript/issues/111

### DIFF
--- a/docs/rules/generic-type-naming.md
+++ b/docs/rules/generic-type-naming.md
@@ -40,5 +40,3 @@ If you do not want to enforce a naming convention for type variables.
 
 ## Further Reading
 - [TypeScript Generics](https://www.typescriptlang.org/docs/handbook/generics.html)
-
-^[A-Z]+$
--- a/docs/rules/generic-type-naming.md
+++ b/docs/rules/generic-type-naming.md
@@ -1,0 +1,42 @@
+# Enforces naming of generic type variables. (generic-type-naming)
+
+It can be helpful to enforce a consistent naming style for generic type variables used within a type.
+For example, prefixing them with  `T`  and ensuring a somewhat descriptive name, or enforcing Hungarian notation.
+
+## Rule Details
+
+This rule allows you to enforce conventions over type variables. By default, it does nothing.
+
+## Options
+
+The rule takes a single string option, which is a regular expression that type variables should match.
+
+Examples of  **correct**  code with a configuration of  `'^T[A-Z][a-zA-Z]+$'`:
+
+```typescript
+type ReadOnly<TType extends object> = {
+  readonly [TKey in keyof TType]: TType[TKey];
+}
+
+interface SimpleMap<TValue> {
+  [key: string]: TValue;
+}
+```
+
+Examples of  **incorrect**  code with a configuration of  `'^T[A-Z][a-zA-Z]+$'`:
+
+```typescript
+type ReadOnly<T extends object> = {
+  readonly [Key in keyof T]: TType[Key];
+}
+
+interface SimpleMap<T> {
+  [key: string]: T;
+}
+```
+
+## When Not To Use It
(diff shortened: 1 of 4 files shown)


docs/rules/generic-type-naming.md
```diff
@@ -40,3 +40,5 @@ If you do not want to enforce a naming convention for type variables.
 
 ## Further Reading
 - [TypeScript Generics](https://www.typescriptlang.org/docs/handbook/generics.html)
+
+^[A-Z]+$
```
j-f1: Is this here on purpose?

unlight: No, by accident

lib/rules/generic-type-naming.js
```diff
@@ -0,0 +1,73 @@
...
+    params.forEach(node => {
+        const type = node.type;
+
+        if (type === "TSTypeParameter" || type === "TypeParameter") {
+            const name = node.name;
+
+            if (name && !regex.test(name)) {
+                const data = { name, rule };
+
+                context.report({ node, messageId: "nameRule", data });
```
macklinu: The `messageId` API was released in [ESLint v4.15.0](https://eslint.org/blog/2018/01/eslint-v4.15.0-released#new-apis). Is it backwards-compatible? I'm wondering if we should use the older API for compatibility with all versions of ESLint:

```js
context.report({ node, message, data })
```

lib/rules/generic-type-naming.js
```diff
@@ -0,0 +1,73 @@
...
+        const rule = context.options[0];
+
+        if (!rule) {
+            return {};
+        }
+        const regex = new RegExp(rule);
+
+        return {
+            VariableDeclarator: node =>
+                checkTypeParameters(node.typeParameters, context, regex, rule),
```
macklinu: I wonder if you could use a higher-order function to DRY up this section:

```js
function createTypeParameterChecker(context, rule) {
  const regex = new RegExp(rule)
  return function checkTypeParameters(node) {
    const params = node.typeParameters && node.typeParameters.params
    // ... rest of logic you already have
  }
}
```

And then your rule would look something like this:

```js
create(context) {
  const rule = context.options[0]
  if (!rule) {
    return
  }

  const checkTypeParameters = createTypeParameterChecker(context, rule)
  return {
    VariableDeclarator: checkTypeParameters, // `node` gets passed to this function
    ClassDeclaration: checkTypeParameters,
    InterfaceDeclaration: checkTypeParameters,
    // etc.
  }
}
```

I like higher-order functions for situations like this, but I wonder if anyone else has feedback on this / if it makes sense. 😄

docs/rules/generic-type-naming.md
```diff
@@ -0,0 +1,42 @@
...
+interface SimpleMap<TValue> {
+  [key: string]: TValue;
+}
+```
+
+Examples of **incorrect** code with a configuration of `'^T[A-Z][a-zA-Z]+$'`:
+
+```typescript
+type ReadOnly<T extends object> = {
+  readonly [Key in keyof T]: TType[Key];
```
macklinu: Should `TType` read `T` in this line?

JamesHenry:
@unlight I think @macklinu has provided some good feedback here. I will give you chance to comment and/or update before merging this.

unlight:
I agree, feedback is good. I'll try to improve code in the coming days.

JamesHenry:
@unlight let me know if you plan on making any other changes, or if this is now ready to go

unlight:
@macklinu Please, take a look. I've implemented your code review notes.
@JamesHenry I'm not planning to make any other changes

JamesHenry:
@unlight @macklinu I am very sorry for the long delay in coming back to this, we are going to work out a better maintenance plan for this project

PR #114, opened 2018-03-11
New: Add no-var-requires

macklinu:
This PR adds the [`no-var-requires`](https://palantir.github.io/tslint/rules/no-var-requires/) rule as listed in the description of #5.

### DIFF
--- a/docs/rules/no-var-requires.md
+++ b/docs/rules/no-var-requires.md
@@ -1,0 +1,28 @@
+# Disallows the use of require statements except in import statements (no-var-requires)
+
+In other words, the use of forms such as `var foo = require("foo")` are banned. Instead use ES6 style imports or `import foo = require("foo")` imports.
+
+## Rule Details
+
+Examples of **incorrect** code for this rule:
+
+```ts
+require("foo");
+var foo = require("foo");
+const foo = require("foo");
+let foo = require("foo");
+```
+
+Examples of **correct** code for this rule:
+
+```ts
+import foo = require("foo");
+```
+
+## When Not To Use It
+
+If you don't care about TypeScript module syntax, then you will not need this rule.
+
+## Compatibility
+
+* TSLint: [no-var-requires](https://palantir.github.io/tslint/rules/no-var-requires/)
--- a/docs/rules/no-var-requires.md
+++ b/docs/rules/no-var-requires.md
@@ -7,7 +7,6 @@ In other words, the use of forms such as `var foo = require("foo")` are banned. 
 Examples of **incorrect** code for this rule:
 
 ```ts
-require("foo");
 var foo = require("foo");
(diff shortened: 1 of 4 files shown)


docs/rules/no-var-requires.md
```diff
@@ -0,0 +1,28 @@
+# Disallows the use of require statements except in import statements (no-var-requires)
+
+In other words, the use of forms such as `var foo = require("foo")` are banned. Instead use ES6 style imports or `import foo = require("foo")` imports.
+
+## Rule Details
+
+Examples of **incorrect** code for this rule:
+
+```ts
+require("foo");
```
j-f1: This should be allowed IMO since there isn’t an equivalent with the `import =` syntax.

tests/lib/rules/no-var-requires.js
```diff
@@ -0,0 +1,66 @@
...
+            errors: [
+                {
+                    message: "Require statement not part of import statement",
+                    line: 1,
+                    column: 11
+                }
+            ]
+        },
+        {
+            code: "require('foo')",
```
macklinu: I believe an unassigned call to `require()` is invalid, but I'm not 100% sure about this based on TSLint's [`no-var-requires` tests](https://github.com/palantir/tslint/blob/master/test/rules/no-var-requires/test.ts.lint). Just calling this out in case anyone has more experience with TSLint's `no-var-requires` rule and if this truly an invalid case.

docs/rules/no-var-requires.md
```diff
@@ -0,0 +1,28 @@
...
+
+Examples of **correct** code for this rule:
+
+```ts
+import foo = require("foo");
+```
+
+## When Not To Use It
+
+If you don't care about TypeScript module syntax, then you will not need this rule.
```
mightyiam: These make me giggle each time.

macklinu: Haha yeah, I agree 😂

docs/rules/no-var-requires.md
macklinu: Yeah, I could see how introducing this as-is would be a breaking change, since all standalone `require()` calls would be flagged by this rule.

I will push up a fix that allows this. If we ultimately want to disallow it before merging, I will just revert the commit. 👍

tests/lib/rules/no-var-requires.js
ajafff: You can replace `require("foo");` with `import "foo";`. Hence an error on this require call is expected.

macklinu: Forgot about that, thanks for pointing that out. I’ll push up a commit next week when I get access to a computer

JamesHenry:
Thanks, @macklinu!

PR #117, opened 2018-03-13
WIP: Add array-type rule

macklinu:
I'd like to contribute the [`array-type`](https://palantir.github.io/tslint/rules/array-type/) rule in #5, but I'm running into issues. None of my `console.log` statements are being hit, so the nodes don't seem to be visited.

Here's an [AST Explorer](https://astexplorer.net/#/gist/5fc8175a8d1d37a145f404d874667300/2a5613d28257fe06656f2649d8434863e1d36676) with the test cases I have, if you'd like to check out the AST.

Why aren't these nodes being visited? Any suggestions on how to approach checking type annotations?


bradzacher:
OOOOOOkaaay.
I did some digging.
A lot of digging.
I setup debugging locally and explored how exactly eslint decides to traverse nodes and their children.

I've outlined the problem here: https://github.com/eslint/typescript-eslint-parser/issues/532
The TL;DR is that eslint doesn't know about `Identifier`'s `typeAnnotation` property, so it doesn't traverse it.

You'll notice in `no-explicit-any` explicitly defines an `Identifier` selector, which manually traverses `typeAnnotation`: https://github.com/nzakas/eslint-plugin-typescript/blob/master/lib/rules/no-explicit-any.js#L89-L95

So to make this work, you'll have to do the same!

lib/rules/array-type.js
```diff
@@ -0,0 +1,58 @@
...
+                console.log(node); // eslint-disable-line no-console
+                if (arrayType === "generic") {
+                    context.report({
+                        node,
+                        message:
+                            "Array type using 'T[]' is forbidden. Use 'Array<T>' instead."
+                    });
+                }
+            },
+            TSTypeReference(node) {
```
armano2: After updating to parser v21 it should be
```suggestion
            "TSTypeReference Identifier[name='array']"(node) {
```

with that you are going to support all cases

lib/rules/array-type.js
```diff
@@ -0,0 +1,58 @@
...
+        ]
+    },
+    create(context) {
+        const arrayType = context.options[0] || "array";
+        //----------------------------------------------------------------------
+        // Public
+        //----------------------------------------------------------------------
+
+        return {
+            TypeAnnotation: console.log, // eslint-disable-line no-console
```
armano2: this one will be not needed i v21

armano2:
@macklinu are you planning to finish this rule?

armano2:
@bradzacher do you mind if i take this one over?

looks like @macklinu is not interested in finishing it

bradzacher:
@armano2 no response doesn't mean not interested. Just means they're busy on other priorities!

I'm not going to stand in the way of progress (and I'm sure @macklinu would be happy for you to commandeer his work if he has other priorities atm)

armano2:
@bradzacher i put it in wrong words 👍

PR #121, opened 2018-03-23
handle optional annotations in type-annotation-spacing

### LINKED ISSUE #120: type-annotation-spacing with optional types
Is it possible to ignore the optional type marker `?` for the "before" option?

right now if you specify space before, you have to put a space between the characters: `? :`, which i don't like - imo it looks ugly, and also it means you can do this:
```TS
type foo = {
   bar? : baz
}
```

I'd much rather enforce this style:
```TS
type foo = {
   bar ?: baz
}
```

bradzacher:
fixes #120

### DIFF
--- a/lib/rules/type-annotation-spacing.js
+++ b/lib/rules/type-annotation-spacing.js
@@ -79,17 +79,24 @@ module.exports = {
          */
         function checkTypeAnnotationSpacing(typeAnnotation) {
             const nextToken = typeAnnotation;
-            const punctuatorToken = sourceCode.getTokenBefore(nextToken);
-            const previousToken = sourceCode.getTokenBefore(punctuatorToken);
-            const type = punctuatorToken.value;
+            const punctuatorTokenEnd = sourceCode.getTokenBefore(nextToken);
+            let punctuatorTokenStart = punctuatorTokenEnd;
+            let previousToken = sourceCode.getTokenBefore(punctuatorTokenEnd);
+            const type = punctuatorTokenEnd.value;
 
             if (punctuators.indexOf(type) === -1) {
                 return;
             }
 
+            if (type === ":" && previousToken.value === "?") {
+                // shift the start to the ?
+                punctuatorTokenStart = previousToken;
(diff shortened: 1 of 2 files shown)


JamesHenry:
@bradzacher an unbelievable amount of test code added here! How come you appear to modify existing tests which did not contain optional types? i.e. why was adding new tests not sufficient here?

bradzacher:
@JamesHenry I glanced through the existing test cases and noticed that they already covered all of the configuration and usage cases.
I figured that it was better to just take the old tests (that could) and make them optional, rather than trying to cover all cases by recreating brand new tests.

JamesHenry:
When existing tests are changed as part of a PR, we naturally lose the reassurance they give us that there were no regressions specific to that test case, that is all I am concerned about here.

Is the same behaviour still covered by other cases?

bradzacher:
sorry, I hadn't looked at the diff it generated, so I didn't realise
It looks like github has made a huge mess of the diff!

I didn't change the existing tests - I copied them, and changed them to cover optional.

The new tests I added are:
- valid: https://github.com/bradzacher/eslint-plugin-typescript/blob/dfbaa4df29328050fc0dcec3a26090d99333a7a9/tests/lib/rules/type-annotation-spacing.js#L1050-L1835
- invalid: https://github.com/bradzacher/eslint-plugin-typescript/blob/dfbaa4df29328050fc0dcec3a26090d99333a7a9/tests/lib/rules/type-annotation-spacing.js#L4171-L5901

bradzacher:
hi there,

it's been a while since this was opened, what's the status?
do you require changes?

this is quite the blocker for my team properly utilising the plugin.

bradzacher:
I have refactored the test cases so that they are separated from the non-optional cases.
The diff is much cleaner now - it is just +lines, with no changed lines.

If you'd like me to go through and cull some of the cases, please let me know.

JamesHenry:
Thanks, @bradzacher! Now we can be completely confident about it being zero risk

PR #124, opened 2018-03-26
add no-inferrable-types rule

43081j:
Related to #5.

Seems all fine, but I only discovered this repo today (and eslint's internals) so do let me know if anything's wrong or could be written better.

### DIFF
--- a/docs/rules/no-inferrable-types.md
+++ b/docs/rules/no-inferrable-types.md
@@ -33,6 +33,9 @@ class Foo {
 
 function fn(a = 5, b = true) {
 }
+
+function fn(a: number, b: boolean, c: string) {
+}
 ```
 
 The following are invalid:
--- a/docs/rules/no-inferrable-types.md
+++ b/docs/rules/no-inferrable-types.md
@@ -1,0 +1,83 @@
+# Disallows explicit type declarations for variables or parameters initialized to a number, string, or boolean. (no-inferrable-types)
+
+Explicit types where they can be easily inferred may add unnecessary verbosity.
+
+## Rule Details
+
+This rule disallows explicit type declarations on parameters, variables
+and properties where the type can be easily inferred from its value.
+
+## Options
+
+This rule has an options object:
+
+```json
+{
+    "ignoreProperties": false,
+    "ignoreParameters": false
+}
+```
+
+### Default
+
+When none of the options are truthy, the following patterns are valid:
+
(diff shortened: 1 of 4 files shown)


docs/rules/no-inferrable-types.md
```diff
@@ -0,0 +1,83 @@
...
+```ts
+const foo = 5;
+const bar = true;
+const baz = 'str';
+
+class Foo {
+    prop = 5;
+}
+
+function fn(a = 5, b = true) {
```
j-f1: Is `function fn(a: number, b: string) {}` valid?

43081j:
Yup, will add it to the docs and tests tomorrow.

We only check assignments so normal type annotations get through fine.

docs/rules/no-inferrable-types.md
43081j: Yup, will add it to the docs and tests tomorrow.

We only check assignments so normal type annotations get through fine.

43081j:
docs updated and tests added

43081j:
@j-f1  any chance you can check that this is all fine now?

would be cool to get this merged by someone

lib/rules/no-inferrable-types.js
```diff
@@ -0,0 +1,187 @@
...
+                });
+        }
+
+        /**
+         * Visits properties
+         * @param {ASTNode} node the node to be visited
+         * @returns {void}
+         */
+        function inferrablePropertyVisitor(node) {
+            if (ignoreProperties || node.readonly) {
```
JamesHenry: How come `readonly` is being given special treatment here? It is not mentioned in your docs

43081j: because of Microsoft/TypeScript#14416, this would not be typed as a boolean:

```ts
readonly foo = false;
```

for readonly properties, you need to specify the type explicitly. so this rule would end up suggesting invalid code if we didn't skip readonly properties.

i dont think its worth mentioning in the docs really as its to avoid introducing a bug we would've otherwise introduced. the expected behaviour is what we have

JamesHenry: Ah ok, please could you add a comment referencing that issue? It is really useful context for future readers of the code.

43081j:
@JamesHenry  comment has been added for reference to the readonly issue FYI

PR #125, opened 2018-04-01
Update: allow expression exclusion from return type checks

### LINKED ISSUE #123: explicit-function-return-type options
I like the idea of requiring explicit return types for methods and functions, as it makes the source very clear. However, requiring it on arrow functions seems OTT.

Could we possibly add a parameter to this or an options object to turn it off for arrow functions?

Happy to try a PR if nobody else is up for it...

43081j:
Fixes #123.

Added an option to exclude arrow functions from the explicit return type rule. Seems to be a common exclusion plenty would appreciate.

So code like this can be excluded from the rule:

```ts
arr.filter((item) => item + 1);
```

Personal preference but one could argue it is over the top to require a return type on the above. So figured why not make it a choice.

### DIFF
--- a/docs/rules/explicit-function-return-type.md
+++ b/docs/rules/explicit-function-return-type.md
@@ -63,6 +63,9 @@ The rule accepts an options object with the following properties:
 
 * `allowExpressions` if true, only functions which are part of a declaration will be checked
 
+By default, `allowExpressions: false` is used, meaning all declarations and
+expressions _must_ have a return type.
+
 ### allowExpressions
 
 Examples of **incorrect** code for this rule with `{ allowExpressions: true }`:
--- a/docs/rules/explicit-function-return-type.md
+++ b/docs/rules/explicit-function-return-type.md
@@ -55,15 +55,6 @@ class Test {
         return;
     }
 }
-
-/*eslint explicit-function-return-type: ["error", { allowExpressions: true }]*/
-node.addEventListener('click', () => {});
-
-/*eslint explicit-function-return-type: ["error", { allowExpressions: true }]*/
-node.addEventListener('click', function() {});
-
-/*eslint explicit-function-return-type: ["error", { allowExpressions: true }]*/
-const foo = arr.map(i => i * i);
 ```
 
 ## Options
@@ -71,6 +62,25 @@ const foo = arr.map(i => i * i);
 The rule accepts an options object with the following properties:
 
 * `allowExpressions` if true, only functions which are part of a declaration will be checked
+
+### allowExpressions
+
+Examples of *incorrect* code for this rule with `{ allowExpressions: true }`:
+
+```ts
+function test() {
+}
+```
+
+Examples of *correct* code for this rule with `{ allowExpressions: true }`:
+
+```ts
+node.addEventListener('click', () => {});
+
+node.addEventListener('click', function() {});
+
+const foo = arr.map(i => i * i);
+```
 
 ## When Not To Use It
 
--- a/docs/rules/explicit-function-return-type.md
+++ b/docs/rules/explicit-function-return-type.md
@@ -65,14 +65,14 @@ The rule accepts an options object with the following properties:
 
 ### allowExpressions
 
-Examples of *incorrect* code for this rule with `{ allowExpressions: true }`:
+Examples of **incorrect** code for this rule with `{ allowExpressions: true }`:
 
 ```ts
 function test() {
 }
 ```
 
-Examples of *correct* code for this rule with `{ allowExpressions: true }`:
+Examples of **correct** code for this rule with `{ allowExpressions: true }`:
 
 ```ts
 node.addEventListener('click', () => {});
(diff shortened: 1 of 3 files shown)


mightyiam:
So, how about changing the logic so that the exemption is not specifically for arrow functions but for function expressions that are not in an assignment? Examples here:
https://github.com/nzakas/eslint-plugin-typescript/issues/123#comment-377810612

43081j:
Would a call expression exclusion not solve your problem?

```ts
// expressions in assignments
const foo = () => {};
const foo = function() {};

// declarations
function foo() {}

// expressions in calls
fn(() => {});
addEventListener('x', () => {});
addEventListener('x', function() {});
arr.map(function mapper() {});
```

seems to me an `excludeCallExpressions` does what you want.

mightyiam:
@43081j almost. With the exception of an IIFE:
```ts
(function () {})() // exempt
```

Should an IFFE be exempt, as well? As a means to prevent pollution of global script scope, I can see that it should be exempt. As a means to return a value, though, it perhaps should not be exempt:

```ts
const foo = (function (): string {
  return 'value'
})()
```

Perhaps this can be another option, like `excludeIIFEs`?

So, I don't mind if it is two options, like `excludeCallExpression` and `excludeIIFEs` or one option, like `excludeNonAssigned`. Yet `excludeCallExpression + excludeIIFEs !== excludeNonAssigned` because the meaning is different.

My logic behind `excludeNonAssigned` is that I would like to have the benefit of explicit return type (documentation, mostly), when it is a function that I'm saving a reference to. Otherwise, it will be an "inline" function and it is usually short and does not benefit from an explicit return type.

43081j:
IIFEs are already excluded (there's a test for it too).

but your second example is a little more complex as we would have to walk up the tree until we find a declaration to detect those (we currently only check the immediate parent).

we should try avoid adding a bunch of options i think and keep it simple. so maybe i could just add a walk up the parents until `Program` or one is a declaration?

ill have a look

mightyiam:
By "declaration", you mean variable declaration?

43081j:
a variable declaration, yes.

i think the current implementation is fine for our original use case. if you really need this to work:

```ts
const foo = (function(): string {})();
```

then we can either traverse the tree or add an iife-specific option. one adds complexity, the other opens the door to an ever-growing list of exclusions. ill take a look tonight at traversing the tree.

mightyiam:
I like [this test](https://github.com/nzakas/eslint-plugin-typescript/pull/125/files?short_path=6ad0bbe#diff-8c1e86a0d47269a28b02f2b6f4a5a474R181). But since it is an arrow function expression, the property `allowExpressions` does not seem appropriate to me.

mightyiam:
@43081j thoughts on this?

## Named function declaration
no exclude option.
```ts
function foo (): undefined {}
```

## unnamed function declaration
no exclude option
```ts
function (): undefined {}
```

BTW, this makes no sense to me, because how would it ever be called?

## assigned function expression
no exclude option
```ts
const functionExpression = function (): undefined {}
const namedFunctionExpression = function named (): undefined {}
const arrowFunctionExpression = (): undefined => {}
```

## unassigned function expression
**exclude option**: `excludeUnassignedExpressions`?

I would make it default for this rule to exclude these. In that case, `includeUnassignedExpressions`?
```ts
document.body.addEventListener('mouseover', function () {}) // function expression
document.body.addEventListener('mouseover', function handleMouseover () {}) // named function expression
;['hi', 'there'].forEach(str => console.log(str), 0) // arrow function expression
;(function () {})() // IIFE
;(() => {})() // IIAFE
```

To make the rule simple, only function expressions that are *themselves the right side of the assignment* will count as being assigned function expressions. So here are some examples of what we should not consider assigned function expressions, even though the execution will result in their assignment.

```ts
const foo = (() => {}) // exempt, to make the rule simple
const operation = typeof item === 'string' ? item => item + '-foo' : item => item + 2 // exempt, to make the rule simple
```

43081j:
This is basically the current behaviour, what you describe, but reversed defaults.

The exclusion right now is of anything that isn't part of a declaration.

Part of a declaration means its a function declaration or its a function expression which has a parent that is a variable declarator.

So assuming we reverse the current defaults and name it as suggested:

```ts

// these are declarations, always checked
function foo() {}
const foo = () => {};
const foo = function() {};

// these are expressions not part of a declaration directly
fn(function() {});
fn(() => {});
const foo = [function() {}, () => {}];
const foo = { fn: () => {} };
const foo = { fn: function() { } };
const foo = { fn() { } };
```

43081j:
@nzakas any chance you can take a look at this so we can possibly get it merged?

43081j:
something like this @JamesHenry?

JamesHenry:
@43081j Great, thanks! Please could you just add an explicit note about the default behaviour/setting?

43081j:
should be sorted now @JamesHenry

PR #132, opened 2018-06-05
Fix: prefer-namespace-keyword false positive (fixes #127)

### LINKED ISSUE #127: prefer-namespace-keyword triggers on "declare global {"
### Code:
```ts
declare global {
  interface Window {
    Vue: VueConstructor
  }
}
```

### Error:
```
  1:1  error  Use 'namespace' instead of 'module' to declare custom TypeScript modules  typescript/prefer-namespace-keyword
```

mysticatea:
Fixes #127.

This PR fixes false positive of `prefer-namespace-keyword` token.

The new logic is more robust because it's token-based instead of text-based. That rule reports `module` keyword to suggest a use of `namespace` keyword, so the new logic finds `module` keyword token. Previously, it reported the declaration if `namespace` keyword was not found, even if `module` keyword was not found.

### DIFF
--- a/lib/rules/prefer-namespace-keyword.js
+++ b/lib/rules/prefer-namespace-keyword.js
@@ -28,47 +28,31 @@ module.exports = {
         const sourceCode = context.getSourceCode();
 
         //----------------------------------------------------------------------
-        // Helpers
-        //----------------------------------------------------------------------
-
-        /**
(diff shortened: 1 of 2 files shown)


JamesHenry:
Thank you @mysticatea!

Issue #33, opened 2017-08-03
Improve documentation for `typescript/no-unused-vars`


OliverJAsh:
There are several use cases `typescript/no-unused-vars` does not yet support, such as `type` declarations:

``` ts
type Foo = string;
// unexpected error
const foo: Foo = 'foo';
```

Would it be possible to track the status of this somewhere in the documentation? The docs for `typescript/no-unused-vars` do not make any mention of this.

I would be happy to raise a PR, but I don't know the full details of what is and isn't supported.

weirdpattern:
@JamesHenry We can close this issue now 😃

OliverJAsh:
Actually I don't think we can? My PR added this rule to the list, but it doesn't add more detail as per this issue.

weirdpattern:
@JamesHenry can we consider removing this rule in favor of TypeScript own flags [--noUnusedParameters and --noUnusedLocals]
(https://github.com/Microsoft/TypeScript/wiki/What%27s-new-in-TypeScript#flag-unused-declarations-with---nounusedparameters-and---nounusedlocals)?.

I think this rule is causing a lot of confusion and giving a lot of false/positives. 
If we already get this out of box, then I don't see the point in keeping it... thoughts?

mightyiam:
I think that TypeScript did not include `noUnusedParameters` and `noUnusedLocals` in `strict` because they see them more as style things. I think they might even regret having these options at all, as they are supposed to be linting issues and not type issues.

lostfictions:
I appreciate having a `no-unused-vars` linter rule because I may not necessarily want `tsc` to return an error (one that might, say, fail a CI build) on something that I might consider more of a warning. (And there's unfortunately no way to configure the severity of the TypeScript flags.)

weirdpattern:
This is a very good point. I will continue working on this rule for as long as it's needed.

tenshiemi:
What is the correct setting for this rule? If I'm understanding correctly, it should prevent ESLint from erroneously throwing these errors, but I can't for the life of me to get it to work and since it's not operating as a rule so much as a workaround it's not clear to me what value I should set it to.

weirdpattern:
Hi @tenshiemi, this rule extends (eslint/no-unused-vars)[https://eslint.org/docs/rules/no-unused-vars] to include TypeScript syntax. 

[...]

PR #135, opened 2018-06-22
Fix: Handle extends generics in no-unused-vars

macklinu:
See #33 

I believe this addresses the first and fourth snippets in https://github.com/nzakas/eslint-plugin-typescript/issues/33#issuecomment-399316751. Any thoughts on other test cases to add or things to consider in this fix?

[Here is a link](http://astexplorer.net/#/gist/2e4055c6371619d206d0f39184c6f835/efbf815025efdfdaa8d6edd3880c8c41b56d1d9a) to an AST Explorer of the test case, if that is helpful for review.


JamesHenry:
Thanks again @macklinu!

PR #136, opened 2018-06-23
add type alias to pascal casing rule

loc:
Enforce Pascal casing for type alias identifiers, in addition to the interface and class definitions already being checked.

### DIFF
--- a/docs/rules/class-name-casing.md
+++ b/docs/rules/class-name-casing.md
@@ -22,6 +22,8 @@ var bar = class invalidName {}
 
 interface someInterface {}
 
+type someTypeAlias = number;
+
 ```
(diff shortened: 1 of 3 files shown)


JamesHenry:
@loc I am very sorry for the long delay in coming back to this, we are going to work out a better maintenance plan for this project.

I believe that TSLint doesn't include TypeAliasDeclarations in the rule that this was modelled after. The rule is also called `class-name-casing` which makes it applicability to clases (and yes, interfaces) a little clearer.

Please could we create a new rule for type aliases instead? I feel they are a distinct enough construct that you may want to different rules for them.

Thanks again for contributing!

PR #140, opened 2018-07-02
[FIX] interface-name-prefix failing on interface with number in name

### LINKED ISSUE #139: interface-name-prefix failing on interface named I18n
We have an interface called `I18n`.
`typescript/interface-name-prefix` throws on that interface

it looks like the problem is here:
https://github.com/nzakas/eslint-plugin-typescript/blob/b804a7c/lib/rules/interface-name-prefix.js#L55
```JS
if (first !== "I" || second !== second.toUpperCase()) {
```

as `'1' === '1'.toUpperCase()` if fails.

bradzacher:
Fixes #139

### DIFF
--- a/lib/rules/interface-name-prefix.js
+++ b/lib/rules/interface-name-prefix.js
@@ -44,7 +44,7 @@ module.exports = {
                 return false;
             }
 
-            return !!name.match(/^I[A-Z]/);
+            return /^I[A-Z]/.test(name);
         }
 
         /**
--- a/lib/rules/interface-name-prefix.js
+++ b/lib/rules/interface-name-prefix.js
@@ -43,19 +43,8 @@ module.exports = {
             if (typeof name !== "string") {
                 return false;
             }
-            if (name.length === 0) {
-                return false;
-            }
-            const first = name.charAt(0);
-            const second = name.charAt(1);
 
-            if (second === "") {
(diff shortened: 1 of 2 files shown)


lib/rules/interface-name-prefix.js
```diff
@@ -43,19 +43,8 @@ module.exports = {
...
-            const second = name.charAt(1);
 
-            if (second === "") {
-                return false;
-            }
-            if (first !== "I" || second !== second.toUpperCase()) {
-                return false;
-            }
-            return true;
+            return !!name.match(/^I[A-Z]/);
```
macklinu: I like this solution! Could also use [`RegExp.prototype.test()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test), which returns a boolean, instead of having to cast the output of `match()` to a boolean:

```js
/^I[A-Z]/.test(name)
```

bradzacher: I prefer that!
I always forget regex's have methods..

JamesHenry:
@bradzacher I am very sorry for the long delay in coming back to this, we are going to work out a better maintenance plan for this project

bradzacher:
@JamesHenry happy to help out if you need it.

PR #145, opened 2018-07-18
Add in three setting configurations (fixes #144)

### LINKED ISSUE #144: add a recommended preset

jayands:
This modifies `lib/index.js` to propose three configurations: "loose",
"recommended" and "strict". Recommended was the model for the other two;
"loose" mode turns everything down to warning or off levels, while everything
is an error in "strict" mode.

Of course, I'd really rather recommend basing it on MS style guides, as
I did not do so.

Signed-off-by: Jonathan <jayands.dev@gmail.com>

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -14,6 +14,143 @@ const path = require("path");
 //------------------------------------------------------------------------------
 // Plugin Definition
 //------------------------------------------------------------------------------
+const plugins = ["typescript"];
 
-// import all rules in lib/rules
-module.exports.rules = requireIndex(path.join(__dirname, "rules"));
+module.exports = {
+    configs: {
+        loose: {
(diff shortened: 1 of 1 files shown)


ezpuzz:
nicer to have this for convenience than have nothing.  thanks for this!

JamesHenry:
@jayands I am very sorry for the long delay in coming back to this, we are going to work out a better maintenance plan for this project.

I feel like with the way this plugin is still evolving, it is too early to lock in a recommended settings list. Users can easily reproduce these things themselves and so it would be a convenient starting point, but not something that would have to live in the plugin just yet.

I'm going to close this for now, thanks again for contributing!

jayands:
Glad to contribute

PR #153, opened 2018-08-10
[FIX] [no-unused-vars] address more false positives

brainkim:
Fixes false positives demonstrated in these test cases (type parameter in new expression, type parameter constraints):
```javascript
import { Nullable } from 'nullable';
class A<T> {
    bar: T
}
new A<Nullable>();
```
```javascript
import { Nullable } from 'nullable';
import { SomeOther } from 'other';
function foo<T extends Nullable>() {
}
foo<SomeOther>();
```
```javascript
import { Nullable } from 'nullable';
import { SomeOther } from 'other';
class A<T extends Nullable> {
    bar: T;
}
new A<SomeOther>();
```
```javascript
import { Nullable } from 'nullable';
import { SomeOther } from 'other';
interface A<T extends Nullable> {
    bar: T;
}
const a: A<SomeOther> = {
    foo: "bar"
};
console.log(a);
```

### DIFF
--- a/lib/rules/no-unused-vars.js
+++ b/lib/rules/no-unused-vars.js
@@ -123,7 +123,9 @@ module.exports = {
                     break;
 
                 case "TSTypeParameter": {
-                    markTypeAnnotationAsUsed(annotation.constraint);
+                    if (annotation.constraint) {
+                        markTypeAnnotationAsUsed(annotation.constraint);
+                    }
                     break;
                 }
                 case "TSMappedType": {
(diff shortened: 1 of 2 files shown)


brainkim:
updated to deal with new expressions and class type parameter constraints

brainkim:
Rebased the PR against master. Unfortunately, I decided after a while to stop using this rule b/c I kept running into false positives. Instead, I'm setting "noUnusedLocals" and "noUnusedParameters" in tsconfig to true.

Two additional cases which may or may not have been fixed since I worked on this:

this type in arguments:
```
function foo(this: Bar) {
  return this.type === "bar";
}
```
will throw an error saying `this` is unused.
public/protected/private keywords in constructor arguments:
```
class Foo {
  constructor(public: bar, private: baz) {}
}
```
will throw an error saying `bar` and `baz` are unused.

brainkim:
https://github.com/nzakas/eslint-plugin-typescript/issues/150 <- related issue.
We might want to create a meta-issue for dealing with all no-unused-var false positives.

bradzacher:
unfortunately no-unused-vars is a difficult rule to maintain due to how many cases typescript introduces.

Thanks for your help in catching more edge cases

PR #137, opened 2018-06-26
Updated NodeType of TSTypeLiteral to TSLiteralType for typescript 2.9.x

weikinhuang:
After upgrading to `typescript@2.9.2` and `typescript-eslint-parser@16.0.0` I've encountered an issue with running the `typescript/no-type-alias` rule against this file:

```typescript
// Manually declaring because the official does not export classvalue
declare module 'classnames' {
  // Type definitions for classnames 2.2
  // Project: https://github.com/JedWatson/classnames
  // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
  // TypeScript Version: 2.3
  export type ClassValue = string | number | ClassDictionary | ClassArray | undefined | null | false;

  interface ClassDictionary {
    [id: string]: boolean | undefined | null;
  }

  export interface ClassArray extends Array<ClassValue> { } // eslint-disable-line typescript/no-empty-interface
  type ClassNamesFn = (...classes: ClassValue[]) => string;

  const classNames: ClassNamesFn;

  export default classNames;
}
```

It would crash with:

```
Cannot read property '0' of undefined
TypeError: Cannot read property '0' of undefined
    at getMessage (/foobar/node_modules/eslint-plugin-typescript/lib/rules/no-type-alias.js:194:26)
    at validateTypeAliases (/foobar/node_modules/eslint-plugin-typescript/lib/rules/no-type-alias.js:280:30)
    at node.types.forEach.type (/foobar/node_modules/eslint-plugin-typescript/lib/rules/no-type-alias.js:296:21)
    at Array.forEach (<anonymous>)
    at validateCompositions (/foobar/node_modules/eslint-plugin-typescript/lib/rules/no-type-alias.js:292:24)
    at validateNode (/foobar/node_modules/eslint-plugin-typescript/lib/rules/no-type-alias.js:310:17)
    at VariableDeclaration (/foobar/node_modules/eslint-plugin-typescript/lib/rules/no-type-alias.js:322:21)
    at listeners.(anonymous function).forEach.listener (/foobar/node_modules/eslint/lib/util/safe-emitter.js:47:58)
    at Array.forEach (<anonymous>)
    at Object.emit (/foobar/node_modules/eslint/lib/util/safe-emitter.js:47:38)
```

At which point, I added this changeset, and some logging for the node, I got:

```
{ type: 'TSLiteralType',
  range: [ 388, 393 ],
  loc:
   { start: { line: 7, column: 95 },
     end: { line: 7, column: 100 } },
[...]

PR #169, opened 2018-11-17
Upgrade typescript-eslint-parser to 17.0.1

armano2:
typescript 2.9
typescript-eslint-parser: 17.0.1

solve #137

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,8 +34,8 @@
     "lint-staged": "^6.0.0",
     "mocha": "^4.0.1",
     "prettier": "^1.11.1",
-    "typescript": "~2.8.1",
-    "typescript-eslint-parser": "^15.0.0"
+    "typescript": "~2.9",
+    "typescript-eslint-parser": "^16.0.1"
   },
   "lint-staged": {
     "*.js": [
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -34,8 +34,8 @@
     "lint-staged": "^6.0.0",
     "mocha": "^4.0.1",
     "prettier": "^1.11.1",
-    "typescript": "~2.8.1",
-    "typescript-eslint-parser": "^15.0.0"
+    "typescript": "~2.9",
+    "typescript-eslint-parser": "^16.0.1"
```
macklinu: I checked this out locally to see how far we could increase the version of typescript + the parser before more tests started breaking, and it looks like we can upgrade to the following safely from this commit.

```
"typescript": "~3.0.3",
"typescript-eslint-parser": "^19.0.2"
```

Should we do that in this PR or a follow up? /cc @bradzacher

armano2:
should i go forward with upgrading or wait for this to be merged?

armano2:
typescript-eslint-parser v20 has some breaking changes, i will prepare fix for them

bradzacher:
thanks @armano2!

feel free to branch off your branch so you can continue working whilst you wait for us to catch up.

in regards to the v20 compatibility.
we'll be releasing these changes as a breaking change to v1.0.0, so don't worry about backwards compatibility.

armano2:
ok :>

PR #175, opened 2018-11-19
[FIX] [no-use-before-define] fix top level scope handling

### LINKED ISSUE #142: [no-use-before-define] false positive when exported class used in type definition
```ts
type TArrayOfTest = Array<Test>;

export class Test {}
```
leads to 

```
  1:27  error  'Test' was used before it was defined  typescript/no-use-before-define
```

with 
```json
"typescript/no-use-before-define": [
  "error", 
  {"functions": false, "typedefs": false, "classes": false}
],
```

**Note:**  without `export` it passes fine:
```ts
type TArrayOfTest = Array<Test>;

class Test {}
```

bradzacher:
Fixes #142

The problem was it was failing to accept the case when both the def and the var were in the top-level global/module scope

### DIFF
--- a/lib/rules/no-use-before-define.js
+++ b/lib/rules/no-use-before-define.js
@@ -43,7 +43,7 @@ function parseOptions(options) {
  * @param {Scope} scope - a scope to check
  * @returns {boolean} `true` if the scope is toplevel
  */
-function isToplevelScope(scope) {
+function isTopLevelScope(scope) {
     return scope.type === "module" || scope.type === "global";
 }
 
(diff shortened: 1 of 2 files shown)


lib/rules/no-use-before-define.js
```diff
@@ -39,6 +39,14 @@ function parseOptions(options) {
     return { functions, classes, variables, typedefs };
 }
 
+/**
+ * @param {Scope} scope - a scope to check
+ * @returns {boolean} `true` if the scope is toplevel
+ */
+function isToplevelScope(scope) {
```
weirdpattern: Casing of the method...

```suggestion
function isTopLevelScope(scope) {
```

PR #174, opened 2018-11-17
Migrate typescript-eslint-parser to 21.0.2

armano2:
To make it work this PR has to be merged due to bug in visitor-tree

- [x] ~~https://github.com/eslint/typescript-eslint-parser/pull/554~~
- [x] ~~https://github.com/eslint/typescript-eslint-parser/pull/560~~
- [x] ~~https://github.com/eslint/typescript-eslint-parser/pull/561~~
- [x] ~~https://github.com/eslint/typescript-eslint-parser/pull/562~~
- [x] ~~https://github.com/eslint/typescript-eslint-parser/pull/565~~
- [x] ~~https://github.com/eslint/typescript-eslint-parser/pull/571~~
- [x] ~~New release of [`typescript-eslint-parser`](https://github.com/eslint/typescript-eslint-parser)~~

unblocks #117, #181 PR

fixes: #152, #155, #126, #164, #166, #171, #193, #189, #143, #33

PR #181, opened 2018-11-24
Add `ban-types` rule

armano2:
I ported this rule and tests from tslint, additionally i added some extra tests to check more cases

this rule relies on changes from https://github.com/bradzacher/eslint-plugin-typescript/pull/174

### DIFF
--- a/docs/rules/ban-types.md
+++ b/docs/rules/ban-types.md
@@ -1,0 +1,73 @@
+# Enforces that types will not to be used (ban-types)
+
+Bans specific types from being used. Does not ban the corresponding runtime objects from being used.
+
+## Rule Details
+
+Examples of **incorrect** code for this rule `"String": "Use string instead"`
+
+```ts
+class Foo<F = String> extends Bar<String> implements Baz<String> {
+    constructor (foo: String) {
+    }
+
+    exit () : Array<String> {
+        const foo: String = 1 as String
+    }
+}
+```
+
+Examples of **correct** code for this rule `"String": "Use string instead"`
+
+```ts
+class Foo<F = string> extends Bar<string> implements Baz<string> {
+    constructor (foo: string) {
+    }
+
+    exit () : Array<string> {
+        const foo: string = 1 as string
+    }
+}
+```
+
+## Options
+```CJSON
+{
+    "typescript/ban-types": ["error", {
+        "types": {
+            // report usages of the type using the default error message
+            "Foo": null,
+
+			// add a custom message to help explain why not to use it
+            "Bar": "Don't use bar!",
+
+            // add a custom message, AND tell the plugin how to fix it
+            "String": {
+			    "message": "Use string instead",
+                "fixWith": "string"
+            }
+        }
+    }
+}
+```
+
+### Example
+```json
+{
+    "typescript/ban-types": ["error", {
+        "types": {
+            "Array": null,
+            "Object": "Use {} instead",
+            "String": {
+                "message": "Use string instead",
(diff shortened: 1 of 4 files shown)


armano2:
yup, i was thinking about this, but i decided to keep interface as is in tslint, i can change it :>

fixer will be nice but it will require users to specify to what they want to change it.
how to you want options to look like with fixer?

```json
{
  "types": {
	"String": "string",
        "Foo": null
  }
}
```
maybe we can change interface a little that, when user is not allowed to specify message, but replace string?

armano2:
@bradzacher i updated options to follow first of your suggestions

bradzacher:
great work!

Maybe add a third option which is an object config?
```CJSON
{
  "Foo": null, // no custom error
  "Bar": "Bar is deprecated, there is no direct replacement", // custom error only
  "String": { // custom error and an optional fixer!
    "message": "Use string instead",
    "fixWith": "string" // nullable, in case there is no fix...
  },
}
```

What do you think?

I'm not sure - depends on our intention with this rule.
Is it purely for flagging things that can't be used, or is it about suggesting alternatives for things?

If it's about forcing alternatives, then having the `null`, and string-only options is counter-productive, because it allows people to not see the fixer option (and it makes your code slightly more complex).

If it's purely for flagging, then the 3 option style is best.

armano2:
and schema is going to look like this:

```js
        schema: [
            {
                type: "object",
                properties: {
                    types: {
                        type: "object",
                        additionalProperties: {
                            oneOf: [
                                { type: "string" },
                                { type: "null" },
                                {
                                    type: "object",
                                    properties: {
                                        message: { type: "string" },
                                        fixWith: { type: "string" }
                                    },
                                    additionalProperties: false
                                }
                            ]
                        }
                    }
                },
                additionalProperties: false
            }
        ]
```

armano2:
@bradzacher i added support for all 3 of them :>

lib/rules/ban-types.js
```diff
@@ -0,0 +1,99 @@
...
+                        }
+
+                        context.report({
+                            node,
+                            messageId: "bannedTypeMessage",
+                            data: {
+                                name: node.name,
+                                customMessage
+                            },
+                            fix(fixer) {
```
bradzacher: would it be better to conditionally define the fixer, instead of having the fixer return conditionally?

i.e.
```
{
  fix: fixWith !== null && fixer => fixer.replaceText(node, fixWith)
}
```

docs/rules/ban-types.md
```diff
@@ -0,0 +1,69 @@
...
+    }
+}
+```
+
+## Options
+```json
+{
+    "typescript/ban-types": ["error", {
+        "types": {
+            "String": TYPE
```
bradzacher: if you change the ` ``` ` type to ` ```CJSON `, you can add comments to the json.
Just so everything is together, merge it in with comments.

```CJSON
{
    "typescript/ban-types": ["error", {
        "types": {
            // report usages of the type using the default error message
            "Foo": null,

			// add a custom message to help explain why not to use it
            "Bar": "Don't use bar!",

            // add a custom message, AND tell the plugin how to fix it
            "String": {
			    "message": "Use string instead",
                "fixWith": "string"
            }
        }
    }
}
```

bradzacher:
Almost forgot - don't forget to update the root README.md!
Fix the tests, address the nits and we can merge this

armano2:
this rule relies on changes from #174

and all tests should pass when #174 is merged

PR #187, opened 2018-11-27
Fix class-name-casing for TSAbstractClassDeclaration

armano2:
This PR adds support for TSAbstractClassDeclaration in `class-name-casing` and fixes issue with:
```ts
var ba_zz = class Foo {};
```

ba_zz was reported as not PascalCase even if class name is Foo


armano2:
@weirdpattern suggestion added

weirdpattern:
@armano2 can you do the same for the test? I think you only add it to the rule

armano2:
I don't really see point in this but ok, added

PR #186, opened 2018-11-27
chore: update package dependencies (fixes #184)

weirdpattern:
Updates package dependencies and introduces eslint and typescript-eslint-parser as peer dependencies

Considerations:
1. The PR includes minor lint changes to 3 rules introduced with the newer eslint version
2. The PR bumps all versions to latest except: eslint-plugin-node (peer dependency of eslint-config-eslint), typescript and typescript-eslint-parser (these two being worked in #174)
3. Peer dependencies are kind of loose (at least eslint), we can change that...

PR #192, opened 2018-11-28
[DOCS] Add contributing to README, and disable vscode autoformat

bradzacher:
The vscode autoformatter is completely at odds with prettier, so if you don't have the prettier extension installed, you end up with wild changes.

Additionally if you've got both the beautify and prettier extensions installed, it can cause some serious issues.

Forcing it to be on will probably cause a lot of headaches to anyone trying to contribute unless they get their environment setup perfectly (sorry @weirdpattern for earlier!!)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -6,3 +6,7 @@ indent_size = 4
 trim_trailing_whitespace = true
 end_of_line = lf
 insert_final_newline = true
+
+# yarn / npm uses 2 spaces when it dumps the file after a change
+[package.json]
+indent_size = 2
--- a/README.md
+++ b/README.md
@@ -71,3 +71,34 @@ This guarantees 100% compatibility between the plugin and the parser.
 * [`typescript/prefer-namespace-keyword`](./docs/rules/prefer-namespace-keyword.md) — Require the use of the `namespace` keyword instead of the `module` keyword to declare custom TypeScript modules. (`no-internal-module` from TSLint)
 * [`typescript/type-annotation-spacing`](./docs/rules/type-annotation-spacing.md) — Require consistent spacing around type annotations
 <!-- end rule list -->
+
+## Contributing
+
+If you're using vscode, we recommend you install both the Pretier and ESLint extensions, and turn on format on save.
+```CJSON
+{
+    // Format a file on save. A formatter must be available, the file must not be auto-saved, and editor must not be shutting down.
+    "[javascript]": {
+        "editor.formatOnSave": true
+    },
+}
+```
+
+For a PR to be merged, all tests must pass, the must be no lint errors, and the code must be formatted.
+- `yarn test`
+- `yarn lint`
+- `yarn format`
+There is a commit hook which will will help you follow this.
+Note that travis will also automatically run the checks when you submit your PR.
+
+When adding or updating a rule, you must:
+- Ensure your feature / bug has an issue behind it.
+  - This just makes it easier for people to find information in future, because PRs aren't included in the default issue search.
+- Ensure your changes are covered by tests.
+  - There's no hard and fast rule for how much testing is required, but try to cover as many bases as you can.
+- Ensure your changes are documented in the `docs` folder. We're working to standardise how we document rules, but your docs should:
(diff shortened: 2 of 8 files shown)


.editorconfig
```diff
@@ -6,3 +6,7 @@ indent_size = 4
 trim_trailing_whitespace = true
 end_of_line = lf
 insert_final_newline = true
+
+# yarn / npm uses 2 spaces when it dumps the file after a change
+[package.json]
```
weirdpattern: Given that .eslintrc is a json file, should we include it here too for consistency?

README.md
```diff
@@ -76,3 +76,24 @@ Then configure the rules you want to use under the rules section.
 * [`typescript/prefer-namespace-keyword`](./docs/rules/prefer-namespace-keyword.md) — Require the use of the `namespace` keyword instead of the `module` keyword to declare custom TypeScript modules. (`no-internal-module` from TSLint)
 * [`typescript/type-annotation-spacing`](./docs/rules/type-annotation-spacing.md) — Require consistent spacing around type annotations
 <!-- end rule list -->
+
+## Contributing
+
+For a PR to be merged, all tests must pass, the must be no lint errors, and the code must be formatted.
```
weirdpattern: `there must be`... instead of `the must be`

README.md
```diff
@@ -76,3 +76,24 @@ Then configure the rules you want to use under the rules section.
...
+- `yarn test`
+- `yarn lint`
+- `yarn format`
+There is a commit hook which will will help you follow this.
```
weirdpattern: `will` instead of `will will`

README.md
```diff
@@ -76,3 +76,24 @@ Then configure the rules you want to use under the rules section.
 * [`typescript/prefer-namespace-keyword`](./docs/rules/prefer-namespace-keyword.md) — Require the use of the `namespace` keyword instead of the `module` keyword to declare custom TypeScript modules. (`no-internal-module` from TSLint)
 * [`typescript/type-annotation-spacing`](./docs/rules/type-annotation-spacing.md) — Require consistent spacing around type annotations
 <!-- end rule list -->
+
+## Contributing
+
+For a PR to be merged, all tests must pass, the must be no lint errors, and the code must be formatted.
+- `yarn test`
```
weirdpattern: Technically `yarn test` will run `yarn lint` for you as it has a `pretest` script.
But I'm ok leaving it as is... emphasizes the importance of running both :smile:...

weirdpattern:
One more thing, now that I've merged #186, you might want to add a new rule to the .gitignore file... I'm excluding everything in the .vscode except for specific files (currently just the settings.json). 

I'm OK with inverting the rule to exclude specific files instead... whatever makes it easier to understand...

weirdpattern:
I just thought of something else... what about moving the contributing section from the README.md to its own CONTRIBUTING.md file? it seems to be a common practice and is even suggested in the [insights/community](https://github.com/bradzacher/eslint-plugin-typescript/community) tab

.editorconfig
bradzacher: The rest of the codebase is 4 spaces!
The problem is that when you do yarn / npm install with old versions, they'll reformat the file to 2 spaces.

This is just to try to stop future vscode autoformat on save problems.

weirdpattern: Ah, that makes sense... let's forget about this

README.md
bradzacher: I just changed the config so that doesn't to make it a bit easier to run tests.
Gets a bit annoying when you're developing a rule to have to wait for the lint + docs check as well.

I've changed the both the precommit hook and travis so they explicitly run all of the checks.

(test run with the new config: https://travis-ci.org/bradzacher/eslint-plugin-typescript/jobs/460887254)

bradzacher:
Man github handles rebases weirdly..

I just:
- Moved to separate CONTRIBUTING.md
- Adjusted `yarn test`/precommit/travis so they match the docs better.
  - `yarn test` only runs the tests
  - precommit / travis both now run the lint, format, test and docs checks

Issue #188, opened 2018-11-28
[prefer-namespace-keyword] Crash when using the rule


iwata:
**Repro**
Just set `typescript/prefer-namespace-keyword` rule and execute linter.

```JSON
{
  "rules": {
    "typescript/prefer-namespace-keyword": "error"
  }
}
```

**Expected Result**
No errors

**Actual Result**
It shows an error below.

```
TypeError: Cannot read property 'range' of undefined
    at SourceCode.getFirstToken (./node_modules/eslint/lib/token-store/index.js:266:18)
    at TSModuleDeclaration (./node_modules/eslint-plugin-typescript/lib/rules/prefer-namespace-keyword.js:40:36)
    at listeners.(anonymous function).forEach.listener (./node_modules/eslint/lib/util/safe-emitter.js:45:58)
    at Array.forEach (<anonymous>)
    at Object.emit (./node_modules/eslint/lib/util/safe-emitter.js:45:38)
    at NodeEventGenerator.applySelector (./node_modules/eslint/lib/util/node-event-generator.js:251:26)
    at NodeEventGenerator.applySelectors (./node_modules/eslint/lib/util/node-event-generator.js:280:22)
    at NodeEventGenerator.enterNode (./node_modules/eslint/lib/util/node-event-generator.js:294:14)
    at CodePathAnalyzer.enterNode (./node_modules/eslint/lib/code-path-analysis/code-path-analyzer.js:632:23)
    at nodeQueue.forEach.traversalInfo (./node_modules/eslint/lib/linter.js:860:28)
```

**Versions**

| package                    | version |
| -------------------------- | ------- |
| `eslint-plugin-typescript` | `0.14.0` |
| `typescript-eslint-parser` | `21.0.1` |
| `typescript`               | `3.1.6` |

weirdpattern:
@iwata can you also try this using the below versions (which are the officially supported versions as of today)

| package                    | version |
| -------------------------- | ------- |
| typescript-eslint-parser | v17.0.1 |
| typescript | 2.9 |

iwata:
@weirdpattern 
I tried and a similar error message appeared.

```
TypeError: Cannot read property 'range' of null
    at SourceCode.getTokenBefore (./node_modules/eslint/lib/token-store/index.js:303:18)
    at checkTypeAnnotationSpacing (./node_modules/eslint-plugin-typescript/lib/rules/type-annotation-spacing.js:84:44)
[...]

PR #195, opened 2018-11-29
Fix issue with declare module without body in prefer-namespace-keyword

armano2:
This PR fixes issue #188

modules can have no body.

-----

there was so much weird logic with tokens there, and we don't need it, it should just take `id` and previous token (module/namespace)....


iwata:
When publish this fixed version?

bradzacher:
This is coming as part of v1.0.0
Which is almost finished.

iwata:
I see. Well then I can hardly wait for the release of v1.0.0!

Issue #199, opened 2018-11-30
ESLint error: Cannot find module 'eslint-plugin-typescript/parser'


wolfy1339:
When trying to run ESLint with this plugin, I get an error:
```
./node_modules/.bin/eslint bot/**/*.ts --ext .ts
Error: Cannot find module 'eslint-plugin-typescript/parser'
    at ModuleResolver.resolve (C:\Users\<user>\Desktop\Programming\github-notification-service\node_modules\eslint\lib\util\module-resolver.js:72:19)
    at loadFromDisk (C:\Users\<user>\Desktop\Programming\github-notification-service\node_modules\eslint\lib\config\config-file.js:514:42)
    at Object.load (C:\Users\<user>\Desktop\Programming\github-notification-service\node_modules\eslint\lib\config\config-file.js:564:20)
    at Config.getLocalConfigHierarchy (C:\Users\<user>\Desktop\Programming\github-notification-service\node_modules\eslint\lib\config.js:227:44)
    at Config.getConfigHierarchy (C:\Users\<user>\Desktop\Programming\github-notification-service\node_modules\eslint\lib\config.js:179:43)
    at Config.getConfigVector (C:\Users\<user>\Desktop\Programming\github-notification-service\node_modules\eslint\lib\config.js:286:21)
    at Config.getConfig (C:\Users\<user>\Desktop\Programming\github-notification-service\node_modules\eslint\lib\config.js:329:29)
    at processText (C:\Users\<user>\Desktop\Programming\github-notification-service\node_modules\eslint\lib\cli-engine.js:180:33)
    at processFile (C:\Users\<user>\Desktop\Programming\github-notification-service\node_modules\eslint\lib\cli-engine.js:240:12)
    at fileList.map.fileInfo (C:\Users\<user>\Desktop\Programming\github-notification-service\node_modules\eslint\lib\cli-engine.js:612:40)
```
ESLint version: 5.9.0
eslint-plugin-typescript version: 0.14.0
Node v11

bradzacher:
Hi @wolfy1339,
The readme changes are part of the upcoming 1.0.0 release.
For the current release readme, please see the tag corresponding to the release (0.14.0), or see the readme on npm.

Duplicate of #198
Please use the search next time.

Issue #198, opened 2018-11-29
Incorrect README or parse not published?


donaldpipowitch:
**Repro**

`$ npm install eslint-plugin-typescript`

**Expected Result**

`node_modules/eslint-plugin-typescript/parser.js` exists.

**Actual Result**

`node_modules/eslint-plugin-typescript/parser.js` doesn't exist.

**Additional Info**

The README says:

> Add eslint-plugin-typescript/parser to the parser field.

And it is shown that way in the example below.

**Versions**

| package                    | version |
| -------------------------- | ------- |
| `eslint-plugin-typescript` | `0.14.0` |
| `typescript-eslint-parser` | `X.Y.Z` |
| `typescript`               | `X.Y.Z` |

bradzacher:
Hi @donaldpipowitch,
The readme changes are part of the upcoming 1.0.0 release.
For the current release readme, please see the tag corresponding to the release (https://github.com/bradzacher/eslint-plugin-typescript/tree/0.14.0), or see the readme on npm

donaldpipowitch:
Cool, thank you 😎

PR #200, opened 2018-12-02
add missing parser.js to npmignore

Vrtak-CZ:
In current configuration of `.npmignore` there is ignored `parser.js` file so `parser: 'eslint-plugin-typescript/parser',` not working because is missing in package on npm registry.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -2,4 +2,5 @@
 !lib/**/*
 !docs/**/*
 !package.json
+!parser.js
 !README.md


bradzacher:
The parser configuration is not working because it has not yet been released.
#198, #199, f0d3910b7674619178e58abbc5c91fe554c03d58

Thanks for this change though, I probably would have forgotten and then realised after the first RC release :sweat_smile:

weirdpattern:
Yeah, I was reading the title wrong... 
I approve this too...

Issue #141, opened 2018-07-06
[no-use-before-define] false positive when exported var name corresponds to interface property name


zuzusik:
```js
interface ITest {
  first : boolean;
  second : string;
  third : boolean;
}

let first = () => console.log('first');

export let second = () => console.log('second');

export namespace Third {
  export let third = () => console.log('third');
}
```

leads to 

```
  2:3  error  'first' was used before it was defined   typescript/no-use-before-define
  3:3  error  'second' was used before it was defined  typescript/no-use-before-define
  4:3  error  'third' was used before it was defined   typescript/no-use-before-define
```

Note: removing both exports leads to error for non exported member be gone, so:
```js
interface ITest {
  first : boolean;
  second : string;
  third : boolean;
}

let first = () => console.log('first');
```
passes just fine

PR #202, opened 2018-12-04
Add test case for ticket #141

armano2:
Test case for #141


armano2:
ok, looks like `no-use-before-define` has one issue in tests

armano2:
There is one test case witch is invalid for typescript:

```js
"use strict";
a();
{
  function a() {}
}
```

scopes/hoisting is working a little different than in js

PR #203, opened 2018-12-04
[FEAT][BREAKING][1/2][member-delimiter-style] Better handling for single line

### LINKED ISSUE #92: [member-delimiter-style] allow delimiter to be specified for single lines
I have these type declarations:

```ts
export type ResolutionMap = Map<Config, {
  plugins: Map<string, Entry>
  presets: Map<string, Entry>
}>

const possibleNames: {
  fromFile: boolean
  name: string
}[] = []
```

I'd like to write them like:

```ts
export type ResolutionMap = Map<Config, {plugins: Map<string, Entry>; presets: Map<string, 
Entry>}>

const possibleNames: {fromFile: boolean; name: string}[] = []
```

However the `member-delimiter-style` configuration doesn't allow for single-line specific configuration.

My current options are `'{delimiter: 'none', requireLast: true, ignoreSingleLine: true}`.

I'd like to give this a try if folks think it's a good idea.

bradzacher:
This PR adds better handling for single line delimiters, and better enforces the `requrieLast` option.

Fixes #92

## Single line delimiters

Previously single line delimiters were either checked or not. This meant that if you used
```
{
    delimiter: "none",
    ignoreSingleLine: false,
}
```
then the fixer would break your code by removing the delimiter from single line statements i.e.:
```TS
interface Foo { bar: string, baz: number }
// fixed to this broken line
interface Foo { bar: string baz: number }
```

### Breaking change:
- Removed option `ignoreSingleLine`
- Added option `singleLine: "none" | "semi" | "comma"`

`singleLine` works the same as `delimiter`, except for single line statements only.

## Stricter `requireLast`

Previously `requireLast` did nothing when turned off, and enforced having a delimiter when turned on.

I felt this was inconsistent with the goals of the linter; being a tool to provide consistent coding conventions.

This is the first of two changes to bring the option into line with eslint's [`comma-dangle`](https://eslint.org/docs/rules/comma-dangle#options). (The next change will replace the boolean with an enum; I didn't want to overload this PR).

### Breaking change:
- `requireLast: false` now enforces that there is no delimiter on the last member.

### DIFF
--- a/lib/rules/member-delimiter-style.js
+++ b/lib/rules/member-delimiter-style.js
@@ -19,17 +19,6 @@ const definition = {
     additionalProperties: false,
 };
 
-const errorMessages = {
-    unexpected: {
-        comma: "Unexpected separator (,).",
-        semi: "Unexpected separator (;).",
-    },
-    expected: {
-        comma: "Expected a comma.",
-        semi: "Expected a semicolon.",
-    },
-};
-
 module.exports = {
     meta: {
         docs: {
@@ -40,6 +29,12 @@ module.exports = {
                 "https://github.com/nzakas/eslint-plugin-typescript/blob/master/docs/rules/member-delimiter-style.md",
         },
         fixable: "code",
+        messages: {
+            unexpectedComma: "Unexpected separator (,).",
+            unexpectedSemi: "Unexpected separator (;).",
+            expectedComma: "Expected a comma.",
+            expectedSemi: "Expected a semicolon.",
+        },
         schema: [
             {
                 type: "object",
@@ -104,25 +99,19 @@ module.exports = {
              * @returns {boolean} the resolved value
              */
             function getOption(type) {
-                // eslint-disable-next-line require-jsdoc
-                function getValue() {
-                    if (isSameLine) {
-                        return opts.singleLine === type;
-                    }
-                    return opts.delimiter === type;
-                }
-
-                if (isLast) {
-                    if (opts.requireLast) {
-                        return getValue();
-                    }
+                if (isLast && !opts.requireLast) {
+                    // only turn the option on if its expecting no delimiter for the last member
                     return type === "none";
                 }
-
-                return getValue();
+                if (isSameLine) {
+                    // use single line config
+                    return opts.singleLine === type;
+                }
+                // use normal config
+                return opts.delimiter === type;
             }
 
-            let message;
+            let messageId;
             let missingDelimiter = false;
             const lastToken = sourceCode.getLastToken(member, {
                 includeComments: false,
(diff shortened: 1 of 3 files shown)


lib/rules/member-delimiter-style.js
```diff
@@ -13,11 +14,22 @@ const definition = {
     properties: {
         delimiter: { enum: ["none", "semi", "comma"] },
         requireLast: { type: "boolean" },
-        ignoreSingleLine: { type: "boolean" },
+        singleLine: { enum: ["none", "semi", "comma"] },
     },
     additionalProperties: false,
 };
```
armano2: i was thinking about this rule, and i feel like this will be more consistent with how eslint specifies options.
```
{
  "multiline": {
    delimiter: { enum: ["none", "semi", "comma"] },
    requireLast: { type: "boolean" },
  },
  "singleline": {
    delimiter: { enum: ["none", "semi", "comma"] },
    requireLast: { type: "boolean" },
  }
}
```

armano2:
```ts
const foo: { a: number } = { a: 1 }
```

```ts
interface foo {
   a: number,
}
```

lib/rules/member-delimiter-style.js
```diff
@@ -89,41 +98,61 @@ module.exports = {
...
+
+                if (isLast) {
+                    if (opts.requireLast) {
+                        return getValue();
+                    }
+                    return type === "none";
+                }
+
+                return getValue();
+            }
```
armano2: you can simplify this function and make it a little more readable:
```js
function getOption(type) {
    if (isLast && !opts.requireLast) {
        return type === "none";
    }
    if (isSameLine) {
        return opts.singleLine === type;
    }
    return opts.delimiter === type;
}
```

lib/rules/member-delimiter-style.js
```diff
@@ -13,11 +14,22 @@ const definition = {
...
+const errorMessages = {
+    unexpected: {
+        comma: "Unexpected separator (,).",
+        semi: "Unexpected separator (;).",
+    },
+    expected: {
+        comma: "Expected a comma.",
+        semi: "Expected a semicolon.",
+    },
+};
```
armano2: instead making object like this you should consider using `messageId`
https://eslint.org/docs/developer-guide/working-with-rules#messageids

bradzacher: i should probably read the docs at some point... 😄

lib/rules/member-delimiter-style.js
bradzacher: I think that's a good idea.
I wanted to do a more drastic change to the config in a separate PR (this one already has ~700 changed lines).

Because this is breaking, I am blocking the 1.0.0 release till the second PR is done.

PR #209, opened 2018-12-08
chore: Add eslint-plugin-eslint-plugin to warn about potential issues

armano2:
I did set up most common rules from it, but it will be nice if we will have also:
- require-meta-docs-url
- require-meta-type
- report-message-format

see: https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin

### DIFF
--- a/tests/lib/rules/type-annotation-spacing.js
+++ b/tests/lib/rules/type-annotation-spacing.js
@@ -1041,11 +1041,6 @@ type Bar = Record<keyof Foo, string>
             parser: "typescript-eslint-parser",
         },
         "let resolver: (() => PromiseLike<T>) | PromiseLike<T>;",
-        `
-interface resolve {
-    resolver: (() => PromiseLike<T>) | PromiseLike<T>;
-}
-        `,
     ],
     invalid: [
         {
@@ -4234,11 +4229,6 @@ type Bar = Record<keyof Foo, string>
             ],
             parser: "typescript-eslint-parser",
         },
-        `
-interface resolve {
-    resolver?: (() => PromiseLike<T>) | PromiseLike<T>;
-}
-        `,
     ],
     invalid: [
         {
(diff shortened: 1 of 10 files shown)


tests/lib/rules/type-annotation-spacing.js
```diff
@@ -4234,11 +4229,6 @@ type Bar = Record<keyof Foo, string>
             ],
             parser: "typescript-eslint-parser",
         },
-        `
-interface resolve {
-    resolver?: (() => PromiseLike<T>) | PromiseLike<T>;
-}
-        `,
```
j-f1: Why the removed tests?

armano2: this test is duplicated 3 times

armano2: https://github.com/bradzacher/eslint-plugin-typescript/blob/master/tests/lib/rules/type-annotation-spacing.js#L25
https://github.com/bradzacher/eslint-plugin-typescript/blob/master/tests/lib/rules/type-annotation-spacing.js#L3458
https://github.com/bradzacher/eslint-plugin-typescript/blob/master/tests/lib/rules/type-annotation-spacing.js#L4237

bradzacher:
for `require-meta-docs-url`, we should probably wrap it up in a utility function so we don't have to worry about having the raw URL in every single rule.

Agree with `report-message-format`. 

Agree with `require-meta-type`. But we can turn that on in a separate PR (considering it'll need a change to every single rule).


Would also like (later, as they'll cause too many errors for this PR):
- `prefer-output-null`
- `no-deprecated-context-methods`
- `test-case-shorthand-strings`

I'll raise issues for the later ones to track it

armano2:
that's why i did not enable them :)

bradzacher:
ah. all good then I've raised the issues.

- prefer-output-null #212
- no-deprecated-context-methods #213
- test-case-shorthand-strings #214
- report-message-format #215
- require-meta-type #216
- require-meta-docs-url #217

PR #211, opened 2018-12-11
Add rule array-type

armano2:
This is continuation of #117

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ This guarantees 100% compatibility between the plugin and the parser.
 <!-- Please run `npm run docs` to update this section -->
 <!-- begin rule list -->
 * [`typescript/adjacent-overload-signatures`](./docs/rules/adjacent-overload-signatures.md) — Require that member overloads be consecutive
-* [`typescript/array-type`](./docs/rules/array-type.md) — Requires using either 'T[]' or 'Array' for arrays (`array-type` from TSLint)
+* [`typescript/array-type`](./docs/rules/array-type.md) — Requires using either `T[]` or `Array<T>` for arrays (`array-type` from TSLint)
 * [`typescript/ban-types`](./docs/rules/ban-types.md) — Enforces that types will not to be used (`ban-types` from TSLint)
 * [`typescript/camelcase`](./docs/rules/camelcase.md) — Enforce camelCase naming convention
 * [`typescript/class-name-casing`](./docs/rules/class-name-casing.md) — Require PascalCased class and interface names (`class-name` from TSLint)
--- a/README.md
+++ b/README.md
@@ -49,6 +49,7 @@ This guarantees 100% compatibility between the plugin and the parser.
 <!-- Please run `npm run docs` to update this section -->
 <!-- begin rule list -->
 * [`typescript/adjacent-overload-signatures`](./docs/rules/adjacent-overload-signatures.md) — Require that member overloads be consecutive
+* [`typescript/array-type`](./docs/rules/array-type.md) — Requires using either 'T[]' or 'Array' for arrays (`array-type` from TSLint)
 * [`typescript/ban-types`](./docs/rules/ban-types.md) — Enforces that types will not to be used (`ban-types` from TSLint)
 * [`typescript/camelcase`](./docs/rules/camelcase.md) — Enforce camelCase naming convention
 * [`typescript/class-name-casing`](./docs/rules/class-name-casing.md) — Require PascalCased class and interface names (`class-name` from TSLint)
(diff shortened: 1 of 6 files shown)


armano2:
@bradzacher i falled into small issue while doing this rule.

everything is working fine except autofix, eslint is working correctly when i change begin and end of node at same time (its not fixing childs)

```
type foo = Array<Array<Bar>>
```

autofix is going to be applied only to first Array not to both

armano2:
^ why?

eslint while applying auto fix resolves it for specific range, `[number, number]` but if you provide array of fixers it's going to compute results from all of them and fill gaps from source code.

armano2:
https://github.com/eslint/eslint/issues/11187

tests/lib/rules/array-type.js
```diff
@@ -9,114 +10,796 @@
...
+                    line: 2,
+                    column: 19,
+                },
+            ],
+        },
+    ],
+});
+
+// eslint rule tester is not working with multi-pass
+// https://github.com/eslint/eslint/issues/11187
```
armano2: for time being i added additional tests outside of rule-tester from eslint to check if output is correct for nested cases

lib/rules/array-type.js
```diff
@@ -0,0 +1,234 @@
...
+        fixable: "code",
+        messages: {
+            errorStringGeneric:
+                "Array type using 'T[]' is forbidden. Use 'Array<T>' instead.",
+            errorStringGenericSimple:
+                "Array type using 'T[]' is forbidden for non-simple types. Use 'Array<T>' instead.",
+            errorStringArray:
+                "Array type using 'Array<T>' is forbidden. Use 'T[]' instead.",
+            errorStringArraySimple:
+                "Array type using 'Array<T>' is forbidden for simple types. Use 'T[]' instead.",
```
bradzacher: To make the errors nicer, could you please make use of the `{{placeholders}}`?

docs/rules/array-type.md
```diff
@@ -0,0 +1,33 @@
+# Requires using either 'T[]' or 'Array' for arrays (array-type)
```
bradzacher: nit
```suggestion
# Requires using either 'T[]' or 'Array<T>' for arrays (array-type)
```

README.md
```diff
@@ -49,6 +49,7 @@ This guarantees 100% compatibility between the plugin and the parser.
 <!-- Please run `npm run docs` to update this section -->
 <!-- begin rule list -->
 * [`typescript/adjacent-overload-signatures`](./docs/rules/adjacent-overload-signatures.md) — Require that member overloads be consecutive
+* [`typescript/array-type`](./docs/rules/array-type.md) — Requires using either 'T[]' or 'Array' for arrays (`array-type` from TSLint)
```
bradzacher: nit
```suggestion
* [`typescript/array-type`](./docs/rules/array-type.md) — Requires using either 'T[]' or 'Array<T>' for arrays (`array-type` from TSLint)
```

lib/rules/array-type.js
```diff
@@ -0,0 +1,234 @@
...
+}
+
+//------------------------------------------------------------------------------
+// Rule Definition
+//------------------------------------------------------------------------------
+
+module.exports = {
+    meta: {
+        docs: {
+            description: "Requires using either 'T[]' or 'Array' for arrays",
```
bradzacher: nit
```suggestion
            description: "Requires using either 'T[]' or 'Array<T>' for arrays",
```

lib/rules/array-type.js
```diff
@@ -0,0 +1,234 @@
...
+            extraDescription: [util.tslintRule("array-type")],
+            category: "TypeScript",
+            url:
+                "https://github.com/nzakas/eslint-plugin-typescript/blob/master/docs/rules/array-type.md",
```
bradzacher: nit
```suggestion
                "https://github.com/bradzacher/eslint-plugin-typescript/blob/master/docs/rules/array-type.md",
```

lib/rules/array-type.js
bradzacher: Mmm good catch. If they use a complex type then you'd have to make sure the message is correct too.

Can we handle the default case where it's just a single identifier (I.e. Use Foo from `Foo[]` or `Array<Foo>`), and for all other cases just use `T`?

That should give the majority case nice errors and save us from the ugly.

armano2: ok, if it's simple type i will update message

armano2: ok, i did change for that, in https://github.com/bradzacher/eslint-plugin-typescript/pull/211/commits/de4ca9d936fe20c73ffe6d4c66c19ec87c326520#diff-00f3c0ea3f6c5b8765c5c323ce639e24R126

PR #219, opened 2018-12-11
[FEAT] [indent] Extend the base rule to support TS nodes

### LINKED ISSUE #201: [indent] Extend base rule to support typescript nodes
#174 adds some tests for indent which covers some default eslint cases.

Unfortunately the rule is designed such that it completely ignores non-standard nodes (i.e. all of the new typescript nodes introduced in parser v20+).

Specifically we need to support:
- `TSTypeLiteral`
- `TSInterfaceBody`
- `TSImportEqualsDeclaration`

bradzacher:
Fixes #201 

The [base eslint implementation](https://github.com/eslint/eslint/blob/master/lib/rules/indent.js) purposely ignores nodes it doesn't know about (i.e. our TS nodes).

Because of how the base rule is written, we have to override the implementation entirely.

### DIFF
--- a/lib/rules/indent.js
+++ b/lib/rules/indent.js
@@ -1,0 +1,82 @@
+/**
+ * @fileoverview Rule to flag non-camelcased identifiers
+ * @author Patricio Trevino
+ */
+"use strict";
+
+const baseRule = require("eslint/lib/rules/indent");
+
+//------------------------------------------------------------------------------
+// Rule Definition
+//------------------------------------------------------------------------------
+
+const KNOWN_NODES = new Set([
+    // ts specific nodes we want to support
+    "TSTypeLiteral",
+    "TSPropertySignature",
+]);
+
+module.exports = Object.assign({}, baseRule, {
+    create(context) {
+        const rules = baseRule.create(context);
+
+        /**
+         * Converts from a TSPropertySignature to a Property
+         * @param {Object} node a TSPropertySignature node
+         * @returns {Object} a Property node
+         */
+        function TSPropertySignatureToProperty(node) {
+            return {
+                type: "Property",
+                key: node.key,
+                value: node.typeAnnotation,
+
+                // Property flags
+                computed: false,
+                method: false,
+                kind: "init",
+                // this will stop eslint from interrogating the type literal
+                shorthand: true,
+
+                // location data
+                range: node.range,
+                loc: node.loc,
+            };
+        }
+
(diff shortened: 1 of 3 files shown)


armano2:
can you add:
- `TSAbstractClassProperty` - `abstract`
- `TSAbstractClassDeclaration` - `abstract`
- `TSEnumDeclaration` - `enum`
- `TSEnumMember` - `enum`
- `TSExportAssignment` - `export = foo`

tests/lib/rules/indent.js
```diff
@@ -21,57 +21,390 @@ const ruleTester = new RuleTester({
...
-    },
-    head: {
-        titleTemplate(title) {
-            if (title) {
-                return \`test\`
-            }
-            return 'Title'
+    valid: [],
+    invalid: [
+        /*
```
armano2: i'm not sure if this is intended but most of tests is commented

bradzacher: ugh. yeah i'm a spaz.. i did that to reduce the noise whilst testing and forgot about it.

lib/rules/indent.js
```diff
@@ -0,0 +1,154 @@
...
+            TSInterfaceBody(node) {
+                // transform it to an ObjectExpression
+                return rules["ObjectExpression, ObjectPattern"]({
+                    type: "ObjectExpression",
+                    properties: node.body.map(TSPropertySignatureToProperty),
+
+                    // location data
+                    range: node.range,
+                    loc: node.loc,
+                });
```
armano2: what's with `implements` -> heritage?
```
interface Foo
implements Bar {
   foo();
}

armano2:
i feel like we should just implement indent for more than those types, TS adds lots of new nodes to AST.
and extends eslint AST...

lib/rules/indent.js
bradzacher: what do you mean sorry?

bradzacher:
I'm happy to add as many types as is required!

The base rule actually works very well by itself, as long as the node type is whitelisted.
It handles most things automatically based on the indentation of previous nodes, but if the node isn't whitelisted it'll skip the indentation altogether.

lib/rules/indent.js
armano2: nvm i was thinking at loud

armano2:
- `TSEmptyBodyDeclareFunction`: `declare function h(x: number): number;`

Issue #215, opened 2018-12-11
[LINT] Turn on eslint-plugin-eslint-plugin/report-message-format


bradzacher:
Continuation of #209 

Use the default regex: `"^[A-Z].*\\.$"`

PR #220, opened 2018-12-12
chore: enable report-message-format and fix all messages

armano2:
I also did small cleanup of test messages in `generic-type-naming.js`
https://github.com/bradzacher/eslint-plugin-typescript/pull/220/files#diff-978785a06f9d7e99cbb60013eac7bea4

when you are using `messageId` you are allowed to use data property, than we don't need custom functions

### config:

```json
 "eslint-plugin/report-message-format": ["error", "^[A-Z'{].*[\\.}]$"]
```

fixes: #215


armano2:
we can slowly migrate rule by rule to `messageId` but we should not do so in one big PR

Issue #217, opened 2018-12-11
[LINT] Turn on eslint-plugin-eslint-plugin/require-meta-docs-url


bradzacher:
Continuation of #209 

Should use a utility function so we don't have to have the URL in every single rule def.

armano2:
@bradzacher can you close this one?

PR #224, opened 2018-12-12
chore: update all urls and replace them with utils.metaDocsUrl

armano2:
#217

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -1,6 +1,11 @@
 "use strict";
 
+const version = require("../package.json").version;
+
 exports.tslintRule = name => `\`${name}\` from TSLint`;
+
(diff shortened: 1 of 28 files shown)


lib/util.js
```diff
@@ -1,7 +1,12 @@
 "use strict";
 
+const version = require("../package.json").version;
+
 exports.tslintRule = name => `\`${name}\` from TSLint`;
 
+exports.metaDocsUrl = name =>
+    `https://github.com/bradzacher/eslint-plugin-typescript/blob/${version}/docs/rules/${name}.md`;
```
bradzacher: nice touch for keeping them in sync with the past!

PR #225, opened 2018-12-13
[FEAT] Add rule no-object-literal-type-assertion

armano2:
This PR implements rule [no-object-literal-type-assertion](https://palantir.github.io/tslint/rules/no-object-literal-type-assertion/)

- #5

### DIFF
--- a/docs/rules/no-object-literal-type-assertion.md
+++ b/docs/rules/no-object-literal-type-assertion.md
@@ -1,0 +1,28 @@
+# Forbids an object literal to appear in a type assertion expression (no-object-literal-type-assertion)
+
+Always prefer `const x: T = { ... };` to `const x = { ... } as T;`
+
+## Rule Details
+
+Examples of **incorrect** code for this rule.
+
+```ts
+const x = { ... } as T;
+```
+
+Examples of **correct** code for this rule.
+
+```ts
+const x: T = { ... };
+```
+
+## Options
+```json
+{
+    "typescript/no-object-literal-type-assertion": "error"
+}
+```
+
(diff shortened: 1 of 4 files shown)


lib/rules/no-object-literal-type-assertion.js
```diff
@@ -0,0 +1,72 @@
...
+                "Forbids an object literal to appear in a type assertion expression",
+            extraDescription: [
+                util.tslintRule("no-object-literal-type-assertion"),
+            ],
+            category: "TypeScript",
+            url:
+                "https://github.com/nzakas/eslint-plugin-typescript/blob/master/docs/rules/no-object-literal-type-assertion.md",
+        },
+        messages: {
+            errorMessage:
```
j-f1: How about `unexpectedTypeAssertion`?

docs/rules/no-object-literal-type-assertion.md
```diff
@@ -0,0 +1,28 @@
...
+Examples of **incorrect** code for this rule.
+
+```ts
+const x = { ... } as T;
+```
+
+Examples of **correct** code for this rule.
+
+```ts
+const x: T = { ... };
```
bradzacher: looking at the tests - it's valid to cast to `any` or `unknown`.
Should probably document this here just to be clear.

armano2: doc updated

armano2:
looks like GH checks crashed, https://travis-ci.com/bradzacher/eslint-plugin-typescript/builds/94678838

PR #231, opened 2018-12-16
[FIX] [generic-type-naming] Add rule

armano2:
This PR changes logic behind `generic-type-naming` and makes it more generic
- TSAbstracClassDeclaration and ***ClassExpression is going to be processed

#### Do not merge
- [x] implement rule
- [x] add tests
- [x] add documentation
- [ ] https://github.com/eslint/typescript-eslint-parser/pull/581 has to be merged first


bradzacher (changes requested):
LGTM

I'll mark as request changes until the parser update is released so that we don't accidentally merge it.

PR #232, opened 2018-12-16
[FEAT] [no-this-alias] Add rule

j-f1:


lib/rules/no-this-alias.js
```diff
@@ -0,0 +1,65 @@
...
+            },
+        ],
+    },
+
+    create(context) {
+        const { allowDestructuring = false, allowedNames = [] } =
+            context.options[0] || {};
+
+        return {
+            VariableDeclarator(node) {
```
armano2:
```suggestion
            "VariableDeclarator[kind!='type']"(node) {
```

do you mind adding check if it's not type alias?

armano2:
@j-f1 you have to update `yarn.lock` you are using `eslint-docs@0.2.6` instead of `eslint-docs@0.2.3`

j-f1:
Done @armano2.

lib/rules/no-this-alias.js
```diff
@@ -0,0 +1,65 @@
...
+                if (init.type !== "ThisExpression") return;
+                if (allowDestructuring && node.id.type !== "Identifier") return;
+
+                if (!allowedNames.includes(id.name)) {
+                    context.report({
+                        node: id,
+                        message:
+                            id.type === "Identifier"
+                                ? "Unexpected aliasing of 'this' to local variable"
+                                : "Unexpected aliasing of members of 'this' to local variables",
```
bradzacher: could you please make use of `messageId` instead?

j-f1:
Done @bradzacher

PR #238, opened 2018-12-16
Add rule `prefer-interface`

armano2:


j-f1:
Is `prefer-interface` a better name?

armano2:
hmm i like `prefer-interface`

lib/rules/prefer-interface.js
```diff
@@ -0,0 +1,72 @@
...
+        //----------------------------------------------------------------------
+        return {
+            // VariableDeclaration with kind type has only one VariableDeclarator
+            "VariableDeclaration[kind='type'] > VariableDeclarator[init.type='TSTypeLiteral']"(
+                node
+            ) {
+                context.report({
+                    node,
+                    messageId: "interfaceOverType",
+                    fix(fixer) {
```
j-f1: Do you think this would be easier to read as a generator?

```js
function fix*(fixer) {
  const typeNode = node.typeParameters || node.id;
  yield fixer.replaceText(sourceCode.getFirstToken(node.parent), "interface");
  yield fixer.replaceTextRange([typeNode.range[1], node.init.range[0]], " ");
  const afterToken = sourceCode.getTokenAfter(node.init);
  if (
    afterToken &&
    afterToken.type === "Punctuator" &&
    afterToken.value === ";"
  ) {
    yield fixer.remove(afterToken);
  }
}
```

tests/lib/rules/prefer-interface.js
```diff
@@ -0,0 +1,87 @@
...
+ruleTester.run("interface-over-type-literal", rule, {
+    valid: [
+        `var foo = { };`,
+        `type U = string;`,
+        `type V = { x: number; } | { y: string; };`,
+        `
+type Record<T, U> = {
+    [K in T]: U;
+}
+        `,
```
j-f1: Should we `.trim()` these? I’ve been doing it for my PRs, but I don’t really mind either way.

j-f1: I really wish we could use something similar to what Prettier has here — plain `.ts` files with the input code, and a snapshot file containing all the error messages.

armano2: we can but it's not following eslint guideline

j-f1: yeah but snapshotting might be a better solution :)

armano2: we can go for it, but not in this PR xd

lib/rules/prefer-interface.js
armano2: its not going to work, `eslint-plugin/fixer-return` is not allowing us to use generator

j-f1: (⏩ https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/issues/74)

tests/lib/rules/prefer-interface.js
armano2: than you should look on this; https://github.com/bradzacher/eslint-plugin-typescript/blob/master/tests/lib/rules/array-type.js#L783

j-f1: (⏩ #242)

PR #239, opened 2018-12-16
[CHORE] Update devDep eslint-plugin-node to 8.0.0

armano2:


armano2:
there is no need to add new rules, we are using recommended config, and there was update for that to

PR #245, opened 2018-12-17
[FEAT] [no-semantic-errors] Add rule (Requires Type Info)

armano2:
This PR replaces need for [`fork-ts-checker-webpack-plugin` ](https://github.com/Realytics/fork-ts-checker-webpack-plugin) while running build with webpack.

Errors reported by this rule comes from typescript Diagnostic services.

- [ ] https://github.com/eslint/typescript-eslint-parser/pull/568 has to be merged first
- [x] I’ve filled out the documentation, including all relevant sections.
- [x] I’ve filled out the rule’s category
- [x] I’ve added the rule documentation link via the helper from `../utils`.

- [x] I’ve added tests for my changes

This rule should not be recommended because its useful only in specific cases.


lib/rules/no-semantic-errors.js
```diff
@@ -0,0 +1,57 @@
...
+    },
+
+    create(context) {
+        const sourceCode = context.getSourceCode();
+
+        const program =
+            context.parserServices && context.parserServices.program;
+
+        if (!program) {
+            return [];
```
armano2: should we print error if there is no `parserServices`?

j-f1: I don’t think so. If the user has a mixed TS and JS codebase, they might have some files without `parserServices`. I might be wrong, though.

armano2: parserServices is available for js files if you are parsing it with eslint-plugin-typescript

bradzacher: Depends on what the likelihood of there not being parserServices?
Isn't that required by the parser, and thus a mandatory thing?

armano2: There is few cases when parserServices is not present:
- there are missing fields in configuration
   ```js
   { 
         generateServices: true,
         project: "./tsconfig.json",
         tsconfigRootDir: __dirname // this field is optional
   }
   ```
- you are using wrong parser (eg. `babel-eslint`)

bradzacher: Well considering that we're going to build more rules based on type information, I would make a utility function to fetch the services from the context, which throws if there is no services.

```TS
const program = getParserServices(context);
//////
function getParserServices(context) {
  if (context.parserServices == null) {
    throw new Error('the world is on fire');
  }
  return context.parserServices;
}
```

armano2: i like this idea, but i will do few more check there

lib/rules/no-semantic-errors.js
```diff
@@ -24,12 +25,7 @@ module.exports = {
     create(context) {
         const sourceCode = context.getSourceCode();
 
-        const program =
-            context.parserServices && context.parserServices.program;
-
-        if (!program) {
-            return [];
-        }
+        const program = util.getParserServices(context).program;
```
j-f1: Would just `util.getProgram(context)` be a more ergonomic API?

armano2: not really we need access to all 3 properties like here: https://github.com/bradzacher/eslint-plugin-typescript/pull/230/files#diff-dc0404c0ba0a0d7ea3cddcb1e786200fR34 or here https://github.com/bradzacher/eslint-plugin-typescript/pull/250/files#diff-a1b523ce3e593ac75a3d6205533c3022R43

armano2: but we could always specify multiple helpers for this `util.getProgram(context)`, `util.getTSNodeFromESTreeNode(context, node)`, `util.getESNodeFromTSTreeNode(context, node)`

hmm...

armano2:
This rule is no longer relevant, as there is going to be support for semantic/synthetic erros in parser/estree 
https://github.com/JamesHenry/typescript-estree/pull/57

Issue #216, opened 2018-12-11
[LINT] Turn on eslint-plugin-eslint-plugin/require-meta-type


bradzacher:
Continuation of #209

PR #247, opened 2018-12-18
[CHORE] enable rule eslint-plugin/require-meta-type

armano2:
This PR enables `eslint-plugin/require-meta-type` rule, and adds type field to all existing rules.

fixes: #216


j-f1:
Is there a list of the valid types and their meanings somewhere?

armano2:
@j-f1 You can find them in eslint documentation https://eslint.org/docs/developer-guide/working-with-rules#rule-basics 

> type (string) indicates the type of rule, which is one of "problem", "suggestion", or "layout":
> - "problem" means the rule is identifying code that either will cause an error or may cause a confusing behavior. Developers should consider this a high priority to resolve.
> - "suggestion" means the rule is identifying something that could be done in a better way but no errors will occur if the code isn’t changed.
> - "layout" means the rule cares primarily about whitespace, semicolons, commas, and parentheses, all the parts of the program that determine how the code looks rather than how it executes. These rules work on parts of the code that aren’t specified in the AST.

j-f1:
> i was thinking about generating additional config base on "layout" type (config-no-layout) to for prettier

I’m planning to submit a PR to [`eslint-config-prettier`](https://github.com/prettier/eslint-config-prettier) to add a `prettier/typescript` config to disable all formatting-related rules.

lib/rules/no-use-before-define.js
```diff
@@ -180,6 +180,8 @@ function isInInitializer(variable, reference) {
 
 module.exports = {
     meta: {
+        type: "problem",
+
```
bradzacher: nit: newline

armano2:
@bradzacher suggestions applied

Issue #249, opened 2018-12-18
Eslint-docs and prettier seem to disagree


uniqueiniquity:
I'm trying to commit some work I've done on implementing a version of TSLint's `no-for-in-array`. However, upon syncing with `master`, it seems like I can't get past the test hook. Specifically, `yarn docs` keeps deleting the top of `indent.md` and `type-annotation-spacing.md`. They also seem to disagree on whether there should be line breaks in the table for the README. It seems like `eslint-docs` is not behaving correctly, but I wanted to file an issue before overriding the test hook. Which should win, `eslint-docs` or `prettier`?

armano2:
try installing new version of packages :>

uniqueiniquity:
Didn't help 😞 Even tried deleting my `node_modules` and starting from scratch.

uniqueiniquity:
For example (the `indent` one is much longer but similar):
```
× The description for type-annotation-spacing must match the required format:
  --- docs\rules\type-annotation-spacing.md
  +++ generated
  @@ -1,10 +1,5 @@
  -# Require consistent spacing around type annotations (type-annotation-spacing)
  -
  -Spacing around type annotations improves readability of the code. Although the most commonly used style guideline for type annotations in TypeScript prescribes adding a space after the colon, but not before it, it is subjective to the preferences of a project. For example:
  -
  -<!-- prettier-ignore -->
  -```ts
  +# Require consistent spacing around type annotations (type-annotation-spacing)
   // with space after, but not before (default if no option is specified)
   let foo: string = "bar";

   // with no spaces
```

j-f1:
Very strange. What OS are you on? Does running `yarn add --dev eslint-docs@latest prettier@latest` fix the issue?

uniqueiniquity:
I'm on Windows (I've had issues with that in the past...).
Will try the command now.

uniqueiniquity:
The command did not fix the issue.

armano2:
what version of node are you using?

uniqueiniquity:
v9.11.1

armano2:
can you send here your modified `type-annotation-spacing.md` -> i'm unable to reproduce this

uniqueiniquity:
[mangled-docs.txt](https://github.com/bradzacher/eslint-plugin-typescript/files/2692462/mangled-docs.txt)

armano2:
[...]

PR #252, opened 2018-12-18
[CHORE] Fix issue with line ending chars in .gitattributes

armano2:
When user has `git core.autocrlf` set to crlf on windows machines there are collisions between `eslint-docs` and `prettier` due to inconsistent line endings in md files.

fixes: #249

@uniqueiniquity

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -6,3 +6,6 @@
 *.js text eol=lf
 *.jsx text eol=lf
 *.d.ts text eol=lf
+*.md text eol=lf
+*.yml text eol=lf
+*.json text eol=lf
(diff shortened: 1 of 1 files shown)


armano2:
@weirdpattern we can do it, i don't see any binary files in project

weirdpattern:
I think the `text` part excludes binary, for binaries you can use `* binary`...

armano2:
> This attribute enables and controls end-of-line normalization. When a text file is normalized, its line endings are converted to LF in the repository.

https://git-scm.com/docs/gitattributes#_checking_out_and_checking_in

Well, basically, it's enforcing that file is going to be read as text file.

weirdpattern:
Still think we should be fine.
As you said before, no binaries just yet in the repo

armano2:
@weirdpattern ok you convinced me, i changed it to `* text eol=lf`

PR #257, opened 2018-12-21
[FIX] Refactor no-namespace rule

armano2:
This PR changes logic in no-namespace rule and fixing some annoying bugs

- improve definition file detection
- migrate messages to messageId
- correct reporting `namespace Foo.Bar {}`
- add unit tests for isDefinitionFile & isTypescript
- add tests for allowDefinitionFiles
- [x] I’ve added tests for my changes

### DIFF
--- a/lib/rules/no-namespace.js
+++ b/lib/rules/no-namespace.js
@@ -20,6 +20,10 @@ module.exports = {
             category: "TypeScript",
             url: util.metaDocsUrl("no-namespace"),
         },
+        messages: {
+            moduleSyntaxIsPreferred:
+                "ES2015 module syntax is preferred over custom TypeScript modules and namespaces.",
+        },
         schema: [
             {
                 type: "object",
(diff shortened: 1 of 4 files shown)


lib/rules/no-namespace.js
```diff
@@ -37,70 +41,28 @@ module.exports = {
...
-
-        /**
-         * Determines if node is a module/namespace declaration.
-         * @param {ASTNode} node the node to be evaluated.
-         * @returns {boolean} true when dealing with declarations.
-         * @private
-         */
-        function isDeclaration(node) {
-            return (
-                node.declare === true && !isTypeScriptModuleDeclaration(node)
```
armano2: this is no longer needed we have `node.global` :)

lib/rules/no-namespace.js
```diff
@@ -37,70 +41,28 @@ module.exports = {
...
         return {
-            TSModuleDeclaration(node) {
+            "TSModuleDeclaration[global!=true][id.type='Identifier']"(node) {
                 if (
-                    node.global ||
-                    isTypeScriptModuleDeclaration(node) ||
-                    (allowDefinitionFiles && isDefinitionFile()) ||
-                    (allowDeclarations && isDeclaration(node))
+                    (node.parent &&
+                        node.parent.type === "TSModuleDeclaration") ||
```
armano2: this is fix for `namespace Foo.Bar {}`

armano2: this is fix for namespace Foo.Bar {}, it was reported 2 times

PR #260, opened 2018-12-21
[FEAT] [BREAKING] [no-unused-vars] remove implicit dep on base rule

bradzacher:
Similar to what we've done with both indent, and camelcase.
Rather than forcing users to configure both our rule and the base rule (which has been a source of confusion), this lets users just use our rule.

```diff
 {
-    "no-unused-vars": ["error", { config }],
+    "no-unused-vars": "off",
-    "typescript/no-unused-vars": "error",
+    "typescript/no-unused-vars": ["error", { config }],
 }
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,7 +85,7 @@ This guarantees 100% compatibility between the plugin and the parser.
 | [`typescript/no-this-alias`](./docs/rules/no-this-alias.md) | Disallow aliasing `this` (`no-this-assignment` from TSLint) |  |  |
 | [`typescript/no-triple-slash-reference`](./docs/rules/no-triple-slash-reference.md) | Disallow `/// <reference path="" />` comments (`no-reference` from TSLint) |  |  |
 | [`typescript/no-type-alias`](./docs/rules/no-type-alias.md) | Disallow the use of type aliases (`interface-over-type-literal` from TSLint) |  |  |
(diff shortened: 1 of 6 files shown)


j-f1:
I’m wondering if we should be copying ESLint’s docs or just linking to them. I’m in favor of linking to the original docs and describing how our rule is different. This allows us to have the most up-to-date docs for the rule that we’re building on, and makes it easier for folks to see how our rule differs.

armano2:
It's not really good idea to do it for no unused vars, other plugins can modify eslint used property, and there can be code effects when using this with eslint-plugin-vue

armano2:
We shouldn't enable eslint rules anyway..

bradzacher:
@armano2 - this won't break eslint-plugin-vue.
All of the eslint plugins run and interop on the same contexts.
It means that if someone is using `eslint-plugin-vue/no-unused-vars`, then it'll still work fine, they'll just be using our rule as their base one instead of the base eslint rule:

```diff
{
    "no-unused-vars": "off",
    "typescript/no-unused-vars": ["error", { config }],
    "vue/no-unused-vars": "error"
}
```

I think in this case it's fine because we are a language-level plugin.
This gives us the freedom to (if we want) switch to use parser services in future (do the parser services report on unused?), without having to do a breaking version bump.

---

@j-f1 I actually thought the same thing. I copied it because that's what we'd done for the other rules.
I'll raise a separate PR to rewrite all of docs together.

armano2:
fair enough i didn't think about 
```
   --noUnusedLocals                    Report Errors on Unused Locals.
   --noUnusedParameters                Report Errors on Unused Parameters.
```

PR #261, opened 2018-12-21
[FEAT] Recommended config

### LINKED ISSUE #144: add a recommended preset

bradzacher:
Fixes #144
Requires ~~#259~~, ~~#260~~.

- added a util to make it standardised and easier to add default config for a rule
- configured recommended based on #144
  - purposely switched `recommended` prop to be `"error" | "warning" | false`
    - inside the eslint repo, it should be `true`. otherwise it's just a property that isn't used officially by eslint. It's truthy so `eslint-docs` still work.
  - changed recommended generator to accept `"error"`/`"warning"` for more configurability.
- adjusted default config of certain rules that didn't match our recommendations.

### DIFF
--- a/lib/configs/recommended.json
+++ b/lib/configs/recommended.json
@@ -33,6 +33,7 @@
         "typescript/no-triple-slash-reference": "error",
         "no-unused-vars": "off",
         "typescript/no-unused-vars": "warning",
-        "typescript/no-use-before-define": "error"
+        "typescript/no-use-before-define": "error",
+        "typescript/no-var-requires": "error"
     }
 }
--- a/lib/configs/recommended.json
+++ b/lib/configs/recommended.json
@@ -18,6 +18,7 @@
         "indent": "off",
         "typescript/indent": "error",
         "typescript/interface-name-prefix": "error",
+        "typescript/member-delimiter-style": "error",
         "typescript/no-unused-vars": true
     }
 }
--- a/lib/configs/recommended.json
+++ b/lib/configs/recommended.json
@@ -31,6 +31,7 @@
         "typescript/no-object-literal-type-assertion": "error",
         "typescript/no-parameter-properties": "error",
         "typescript/no-triple-slash-reference": "error",
-        "typescript/no-unused-vars": true
+        "no-unused-vars": "off",
+        "typescript/no-unused-vars": "warning"
     }
 }
--- a/lib/configs/recommended.json
+++ b/lib/configs/recommended.json
@@ -35,6 +35,7 @@
         "typescript/no-unused-vars": "warning",
         "typescript/no-use-before-define": "error",
         "typescript/no-var-requires": "error",
-        "typescript/prefer-interface": "error"
+        "typescript/prefer-interface": "error",
+        "typescript/prefer-namespace-keyword": "error"
     }
 }
--- a/lib/configs/recommended.json
+++ b/lib/configs/recommended.json
@@ -1,6 +1,22 @@
 {
+    "parser": "eslint-plugin-typescript/parser",
+    "parserOptions": {
+        "sourceType": "module"
+    },
+    "plugins": [
+        "typescript"
+    ],
     "rules": {
-        "class-name-casing": "error",
-        "no-unused-vars": "error"
+        "typescript/adjacent-overload-signatures": "error",
+        "typescript/array-type": "error",
+        "typescript/ban-types": "error",
+        "camelcase": "off",
+        "typescript/camelcase": "error",
+        "typescript/class-name-casing": "error",
+        "typescript/explicit-function-return-type": "warning",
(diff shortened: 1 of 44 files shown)


lib/rules/adjacent-overload-signatures.js
```diff
@@ -18,6 +18,7 @@ module.exports = {
             category: "TypeScript",
             extraDescription: [util.tslintRule("adjacent-overload-signatures")],
             url: util.metaDocsUrl("adjacent-overload-signatures"),
+            recommended: "error",
```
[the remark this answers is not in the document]
bradzacher: if you're inside the eslint repo, yes.
otherwise it doesn't matter.

bradzacher:
@Aladdin-ADD - I have [updated the readme](https://github.com/bradzacher/eslint-plugin-typescript/blob/recommended/README.md)

armano2:
@bradzacher looks like there are errors in tests for `no-type-alias`

bradzacher:
@armano2 yup - it needed the config changes from #259

arvigeus:
It does not work for me. ESLint complains that "warning" should be "warn" instead. Bonus question: how to make it work with prettier?

aladdin-add:
it was fixed in master, but not released yet.

@bradzacher plan to release a patch version?

bradzacher:
I'm back from holidays so I'll look to cleaning things up and releasing `rc.1`

lib/configs/recommended.json
```diff
@@ -1,7 +1,42 @@
...
+        "typescript/adjacent-overload-signatures": "error",
+        "typescript/array-type": "error",
+        "typescript/ban-types": "error",
+        "camelcase": "off",
+        "typescript/camelcase": "error",
+        "typescript/class-name-casing": "error",
+        "typescript/explicit-function-return-type": "warning",
+        "typescript/explicit-member-accessibility": "error",
+        "indent": "off",
+        "typescript/indent": "error",
```
aladdin-add: `recommended` was designed to only contain possible errors and best practices rules, the `typescript/indent` as a stylistic rule, I'm not sure if it could be included in the recommended.

arvigeus: I agree. Although it seems like all `remommended` rulesets out there include stylistic rules as well, probably for the folks who don't use `prettier`. I think the best solution would be to leave these rules in `recommended`, and add `plugin:typescript/prettier` to shut them off. Thoughts?
I also noticed other differences compared to `tslint:recommended`, for example complaining about using `IFoo` as interface name instead of `Foo`. Correct me if I am wrong, but `IFoo` seems like the more popular convention, although on typescript docs site I don't see them used.

j-f1: There’s an `eslint-config-prettier`, which I’m going to add a `typescript` preset to (https://github.com/prettier/eslint-config-prettier/pull/68), which will disable all stylistic rules. People could then extend `["plugin:typescript/recommended", "prettier/typescript"]`.

armano2: we should not enable stylistic rules as recommended, they are opinionated, some ppl prefer tabs over spaces xd

bradzacher: in particular for the second point. There are three possible cases:
1) Don't disable `indent`, don't turn on `typescript/indent`.
    - If they apply the airbnb config, then ours, and we don't disable `indent`, then they'll get a number of incorrect indentation errors due to the incorrect support of TS nodes. This could possibly lead to users opening issues because of the incorrect handling from the base rule.
2) Disable `indent`, don't turn on `typescript/indent`.
    - This is terrible because it means the user will lose their indentation errors, and they mightn't even know we disabled them.
3) Disable `indent`, turn on `typescript/indent`.
    - Stops users using the broken rule, and starts flagging errors. User knows the rule is misconfigured for their code, and adjusts accordingly.

[Looking at some stats](https://ukupat.github.io/tabs-or-spaces/) - of the top 292 starred TS repos; 45% use 4 spaces, 36% use 2 spaces, and 16% use tabs (0.68% uses 3 spaces - wtf).
So with this setting we're likely targeting the most common use case.

bradzacher: I don't have any data, but my gut would say that the majority use case is TS only. For those that have a mixed codebase, they're probably advanced enough users that they can apply overrides themselves (not that they need to, `typescript/indent` works fine with plain JS code).

Turning off the base rule is [listed as part of the config in the docs for our indent rule.](https://github.com/bradzacher/eslint-plugin-typescript/blob/master/docs/rules/indent.md#options).

I don't like the idea of only disabling `indent`, whether it be on everything, or just TS(X) files, because unless a user interrogates the recommended config, they won't know we disabled it, and could then fall into using inconsistent indent when they thought their linter was working.

armano2: what's about vue files? more and more ppl is moving to typescript :)
vue + ts + eslint

PR #270, opened 2018-12-28
[FEAT] Add getParserServices function

bradzacher:
@armano2 closed #245 because it won't be needed.

But it had a helpful function in it for us to ensure people are using our parser, so we don't get cryptic error messages about reading properties from undefined.

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -117,6 +117,9 @@ exports.getParserServices = context => {
         !context.parserServices.program ||
         !context.parserServices.esTreeNodeToTSNodeMap
     ) {
+        // TODO - the message will require revisiting once the typescript-eslint-parser/typescript-estree finalises
+        //        their work around exposing the parser. They may require that there be a project config field in
+        //        the eslint config, in which case we should check and/or report that here appropriately.
         throw new Error(
             "This rule requires you to use `eslint-plugin-typescript/parser`."
         );
--- a/lib/util.js
+++ b/lib/util.js
@@ -75,7 +75,7 @@ exports.deepMerge = deepMerge;
  * @param {any[]} userOptions the user opts
  * @returns {TOptions} the options with defaults
  */
-function applyDefault(defaultOptions, userOptions) {
+exports.applyDefault = (defaultOptions, userOptions) => {
     // clone defaults
     const options = JSON.parse(JSON.stringify(defaultOptions));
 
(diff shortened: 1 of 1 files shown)


lib/util.js
```diff
@@ -97,15 +97,29 @@ function applyDefault(defaultOptions, userOptions) {
...
+ * @returns {{esTreeNodeToTSNodeMap}|{program}|Object|*} parserServices
+ */
+exports.getParserServices = context => {
+    if (
+        !context.parserServices ||
+        !context.parserServices.program ||
+        !context.parserServices.esTreeNodeToTSNodeMap
+    ) {
+        throw new Error(
+            "This rule requires you to use `typescript-eslint-parser`."
```
j-f1:
```suggestion
            "This rule requires you to use `eslint-plugin-typescript/parser`."
```

armano2: ithink this message is wrong anyway, it requires project field in eslint parser config.

bradzacher: does the user have to provide the project field?
it seems like a fiddly bit of config if it can just use the app root...

bradzacher: hmm we'll have to figure it out once they have finalised everything.

I hope we can make that an optional thing, and have it use the root folder by default.
If not we will have to write up some decent docs on how to configure it.

uniqueiniquity: @bradzacher I just noticed this comment. Is there a different way you'd prefer the semantic rules to be configured? My original intent was to match tslint, where a tsconfig has to be provided in order to power the rules. But if there's a more implicit way that makes more sense, I'd be happy to try to work it in.

bradzacher: Ideally I'd like to have it implicitly using the root tsconfig unless otherwise specified. This would make everything just work out of the box. 

Unfortunately right now if you want to use the plugin, you can; but if you want to use an upcoming rule that requires type info, you have to point it at your tsconfig.

This creates a confusing dichotomy in the configuration where it can both work and not work at the same time.

Additionally in my experience, explicitly defining paths in config files can sometimes be confusing depending on project folder structure, and what folder the tool considers the root.

So if we can make it work for the default case, then it'll be easy for users (which means less question type issues being opened).

uniqueiniquity: Got it. I should mention that the concern here is that the information we're providing through parser services incurs a major performance hit (as in TSLint). So I tried to make it as opt-in as possible so as to not completely sink users' lint times.
I'd be happy to help put together docs to make it as clear as we can, but I would like to maintain the opt-in nature of semantic rules.

PR #274, opened 2019-01-04
[FIX] [indent] Add more indent cases + tests

bradzacher:
- Added many validation tests for our indent rule running on TS nodes.
- Added an automatic generator to test the fixer as well as valid cases.
- Handled more TS-specific node cases


lib/rules/indent.js
```diff
@@ -14,25 +14,68 @@ const util = require("../util");
...
+    "TSAnyKeyword",
+    "TSBooleanKeyword",
+    "TSNeverKeyword",
+    "TSNumberKeyword",
+    "TSStringKeyword",
+    "TSSymbolKeyword",
+    "TSUndefinedKeyword",
+    "TSUnknownKeyword",
+    "TSVoidKeyword",
+    // null is treated as a Literal
```
armano2: thats not true, in current version its kinda broken, and only in some cases you will see `TSNullKeyword`, it got fixed in new release....

https://astexplorer.net/#/gist/de2632feeea6ac57b18f0bdbcf43a01e/7cec8d963907d6839e62c81bdba4f46028938a10

you can see in AST there is `TSNullKeyword`

bradzacher: Ahh see I tried it as a type annotation, and it didn't work. I didn't even think it would work on that side of the equation!

armano2: like i said its broken in `typescript-estree@5.3.0` i solved it in https://github.com/JamesHenry/typescript-estree/pull/65

